PR #6, opened 2026-01-26
fix: `docker-compose` -> `docker compose`

pandeymangg:
`docker-compose` will be deprecated, so we change it now to `docker compose`


xernobyl (approved):
LGTM

PR #14, opened 2026-02-03
feat: Add goose migrations for database schema

xernobyl:
## Summary
Introduces [goose](https://github.com/pressly/goose) for SQL migrations and adds the initial schema plus tooling/CI for running and validating migrations.

## Changes
- **Migrations**: Added `migrations/001_initial_schema.sql` with goose annotations (`-- +goose up` / `-- +goose down`). Defines `feedback_records` table, `field_type_enum`, indexes, and enables `vector` and `pgcrypto` extensions.
- **Makefile**: `make init-db` runs migrations (goose up) using `DATABASE_URL`; `make migrate-status` and `make migrate-validate` for status and validation (no DB). `make install-tools` installs goose; `make dev-setup` runs `init-db` and `install-hooks`.
- **CI**: New workflow `migrations-validate.yml` runs on PRs when `migrations/` or the workflow file change—installs goose and runs `make migrate-validate`. Existing workflows (tests, api-contract-tests) install goose and run `make migrate-validate` where relevant.
- **Pre-commit**: When files under `migrations/` are staged, the hook runs `make migrate-validate` (requires goose; install via `make install-tools`).
- **Docs**: `AGENTS.md`, `README.md`, and `.githooks/README.md` updated to describe goose, the new make targets, and migration workflow.
- **Compose**: `docker-compose.yml` was renamed to `compose.yml`.
- **OpenAPI**: Unused fields were removed from the API spec.

## How to verify
- `make install-tools` then `make migrate-validate` — passes.
- With `DATABASE_URL` set: `make init-db` then `make migrate-status` — migrations apply and status is shown.
- Push and open a PR; migrations-validate job runs when `migrations/` (or the workflow) changes.


migrations/001_initial_schema.sql
```diff
@@ -64,3 +65,9 @@ CREATE INDEX idx_feedback_records_tenant_user_identifier ON feedback_records(ten
 CREATE INDEX idx_feedback_records_tenant_collected_at ON feedback_records(tenant_id, collected_at);
 CREATE INDEX idx_feedback_records_tenant_source_type ON feedback_records(tenant_id, source_type);
 CREATE INDEX idx_feedback_records_tenant_field_type ON feedback_records(tenant_id, field_type);
+
+-- +goose down
+DROP TABLE IF EXISTS feedback_records;
+DROP TYPE IF EXISTS field_type_enum;
+DROP EXTENSION IF EXISTS vector;
+DROP EXTENSION IF EXISTS "pgcrypto";
```
mattinannt: AI Feedback:
In shared database environments, other tables may depend on these extensions. Consider:
- Removing extension drops from down migration
- Adding a comment noting this is only safe for isolated deployments
----
I'm currently thinking we should skip these `DROP EXTENSION` then.

AGENTS.md
```diff
@@ -13,7 +13,7 @@
 - `make build`: build the API binary to `bin/api`.
 - `make tests`: run integration tests in `tests/`.
 - `make tests-coverage`: generate `coverage.html`.
-- `make init-db`: initialize database schema using `DATABASE_URL`.
+- `make init-db`: run goose migrations up using `DATABASE_URL`. `make migrate-status` and `make migrate-validate` for status and validation. New migrations go in `migrations/` with goose annotations.
```
mattinannt: Should we also add an instruction for migration naming convention here? Or is goose handling this?

mattinannt (dismissed):
@xernobyl thank you for the PR :-) Looks good! While reading through the code I only had two suggestions.
Let's see what @BhagyaAmarasinghe finds with a bit more Go & Goose experience :-)

migrations/001_initial_schema.sql
xernobyl: Removed.

AGENTS.md
xernobyl: Migrations are named manually. I added a comment regarding the naming conventions.

BhagyaAmarasinghe (approved):
Looks good 🚀

PR #16, opened 2026-02-05
feat: Webhooks support

xernobyl:
Adds Standard Webhooks support for Hub events, based on River as a queue.

### Event / Webhooks pipeline (architecture)

1. **API request** – Client calls a mutation (e.g. `POST /v1/feedback-records`, `DELETE /v1/webhooks/{id}`). The handler calls the corresponding service.
2. **Service publishes event** – The service persists the change, then calls **MessagePublisherManager** with event type and payload (e.g. `PublishEvent(FeedbackRecordCreated, record)`). This is fire-and-forget into an in-memory channel; the API does not wait for webhook delivery.
3. **MessagePublisherManager fans out** – A background goroutine reads from the channel and calls **PublishEvent** on each registered provider (today: only **WebhookProvider**). Per-event timeout is configurable via `MESSAGE_PUBLISHER_PER_EVENT_TIMEOUT_SECONDS` (default 10s) so one provider doesn’t block others. If the channel is full, the event is dropped and a warning is logged. Channel buffer size is configurable via `MESSAGE_PUBLISHER_BUFFER_SIZE` (default 1024).
4. **WebhookProvider enqueues jobs** – For each event, it queries the DB for enabled webhooks that subscribe to that event type, builds a batch of job params (capped by `WEBHOOK_MAX_FAN_OUT_PER_EVENT`, default 500), and calls River’s **InsertMany** once. No HTTP here; only list + enqueue. Jobs are unique by (event id, webhook id) for 24 hours.
5. **River runs the worker** – River picks up jobs and runs **WebhookDispatchWorker** for each. The worker loads the webhook by ID, builds the payload, and calls **WebhookSender**.
6. **WebhookSender delivers** – Signs the payload (Standard Webhooks), sets headers (`webhook-id`, `webhook-signature`, `webhook-timestamp`), POSTs to the webhook URL. 2xx → job complete. 410 Gone → webhook disabled in DB, no retry. Other failures → retry with backoff until max attempts; after last failure the webhook can be disabled.

**In short:** API → Service → MessagePublisherManager (channel) → WebhookProvider (DB + River InsertMany) → River queue → WebhookDispatchWorker → WebhookSender → HTTP POST. The slow/unreliable part (HTTP to third-party URLs) stays in the worker and sender; the API stays non-blocking.

### What’s in this branch

**Webhooks**
- **CRUD API** for webhook endpoints: `POST/GET/PATCH/DELETE /v1/webhooks`, with list/filter support.
- **Event types**: `feedback_record.created`, `feedback_record.updated`, `feedback_record.deleted` (and webhook lifecycle events).
- **Delivery**: River-backed worker sends signed payloads (Standard Webhooks) to subscriber URLs with retries and configurable concurrency/attempts.
- **Batch enqueue**: Provider uses River’s **InsertMany** (one call per event) and a configurable fan-out cap per event (`WEBHOOK_MAX_FAN_OUT_PER_EVENT`, default 500) to avoid blocking the publisher when an event has many webhooks.
- **Migrations**: `002_webhooks.sql`, `003_webhook_disabled_state.sql` (and dependency on goose migrations).
- **OpenAPI**: New webhook schemas and operations in `openapi.yaml`.

**Infrastructure & config**
- **River**: River client and `WebhookDispatchWorker` in `cmd/api/main.go`; webhook provider enqueues jobs via **InsertMany** (one batch per event, capped by `WEBHOOK_MAX_FAN_OUT_PER_EVENT`).
- **Docker Compose**: `riverui` service for the River UI (port 8081), with basic auth via `RIVER_BASIC_AUTH_USER` / `RIVER_BASIC_AUTH_PASS` (no defaults in compose).
- **Config**: `WEBHOOK_DELIVERY_MAX_CONCURRENT`, `WEBHOOK_DELIVERY_MAX_ATTEMPTS`, `WEBHOOK_MAX_FAN_OUT_PER_EVENT`; `MESSAGE_PUBLISHER_BUFFER_SIZE`, `MESSAGE_PUBLISHER_PER_EVENT_TIMEOUT_SECONDS`, `SHUTDOWN_TIMEOUT_SECONDS`; `.env.example` updated with River UI auth and the above vars.

**Other**
- **Message publisher**: Event publishing abstraction used by the feedback-records service to trigger webhook delivery; buffer size and per-event timeout configurable.
- **Integration tests**: Webhook CRUD and auth covered in `tests/integration_test.go`.

### How to try it
1. `make init-db` and `make river-migrate`.
2. `docker compose up -d` for Postgres + River UI (set River UI auth in `.env`).
3. Create a webhook (e.g. `feedback_record.created`) and create a feedback record; confirm delivery in River UI and at the webhook URL.


internal/service/feedback_records_service.go
```diff
@@ -81,26 +88,70 @@ func (s *FeedbackRecordsService) UpdateFeedbackRecord(ctx context.Context, id uu
...
 	}
 
-	count, err := s.repo.BulkDelete(ctx, userIdentifier, tenantID)
+	records, err := s.repo.BulkDelete(ctx, userIdentifier, tenantID)
 	if err != nil {
 		return 0, fmt.Errorf("bulk delete feedback records: %w", err)
 	}
-	return count, nil
+
+	for i := range records {
```
BhagyaAmarasinghe: Here bulk delete can silently drop events. This loop publishes one event per deleted record. With the default channel buffer of 1024, deleting >1024 records in a single call will start silently dropping events via the `select/default` in `PublishEventWithChangedFields`. Memory also scales linearly since all deleted rows are loaded into memory from DELETE ... RETURNING.

We can publish a single `feedback_record.bulk_deleted` event containing all IDs, or at minimum document this limit and consider a larger buffer / blocking send for bulk operations.

internal/models/webhooks.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return fmt.Errorf("parse event types: %w", err)
+	}
+	w.EventTypes = parsed
+	return nil
+}
+
+// CreateWebhookRequest represents the request to create a webhook.
+type CreateWebhookRequest struct {
+	URL        string                `json:"url" validate:"required,no_null_bytes,min=1,max=2048"`
```
BhagyaAmarasinghe: The validate tag is `required,no_null_bytes,min=1,max=2048` but doesn't check that the value is actually a URL. A user could create a webhook with url: "not a url", which would fail at delivery time and eventually auto-disable. Add http_url or url to the validate tag
URL string `json:"url" validate:"required,no_null_bytes,http_url,min=1,max=2048"

internal/api/handlers/webhooks_handler.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	if err := validation.ValidateStruct(&req); err != nil {
+		validation.RespondValidationError(w, err)
+		return
+	}
+
+	webhook, err := h.service.CreateWebhook(r.Context(), &req)
+	if err != nil {
+		if errors.Is(err, huberrors.ErrNotFound) {
```
BhagyaAmarasinghe: If I am correct, this looks like dead code. A `POST /v1/webhooks` (create) endpoint should never return 404. The service `CreateWebhook` doesn't return `ErrNotFound`. This looks like a paste from the Get/Update/Delete handlers. Please check if this needs to be removed.

internal/service/webhooks_service.go
```diff
@@ -0,0 +1,151 @@
...
+// NewWebhooksService creates a new webhooks service.
+func NewWebhooksService(repo WebhooksRepository, publisher MessagePublisher) *WebhooksService {
+	return &WebhooksService{
+		repo:      repo,
+		publisher: publisher,
+	}
+}
+
+// CreateWebhook creates a new webhook.
+func (s *WebhooksService) CreateWebhook(ctx context.Context, req *models.CreateWebhookRequest) (*models.Webhook, error) {
```
BhagyaAmarasinghe, line 46: Signing key validation is missing at creation. When a user provides a `signing_key`, there's no check that it's valid for Standard Webhooks (base64-decodable, correct prefix/length). If the key is malformed, `standardwebhooks.NewWebhook()` will fail at send time, causing every delivery attempt to fail until the webhook is auto-disabled.

We can validate the user-supplied key here (e.g. try `standardwebhooks.NewWebhook(req.SigningKey)`) before persisting.

migrations/002_webhooks.sql
```diff
@@ -0,0 +1,20 @@
...
+-- Webhooks table for Standard Webhooks implementation
+
+CREATE TABLE webhooks (
+  id UUID PRIMARY KEY DEFAULT uuidv7(),
+  url VARCHAR(2048) NOT NULL,
+  signing_key VARCHAR(255) NOT NULL,
+  enabled BOOLEAN NOT NULL DEFAULT true,
+  tenant_id VARCHAR(255),
+  event_types VARCHAR(64)[],
+  created_at TIMESTAMP NOT NULL DEFAULT NOW(),
```
BhagyaAmarasinghe: There's inconsistent timestamp types. `created_at` and `updated_at` use `TIMESTAMP` (without timezone), but `disabled_at` in migration 003 uses `TIMESTAMPTZ`. Using `TIMESTAMP` without timezone is generally a PostgreSQL anti-pattern, if the server timezone changes, stored values will be interpreted incorrectly. Change to `TIMESTAMPTZ` for consistency:

internal/repository/webhooks_repository.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+	return r.List(ctx, filters)
+}
+
+// maxWebhookListLimit caps the number of webhooks returned for delivery to avoid unbounded queries.
+const maxWebhookListLimit = 1000
+
+// ListEnabledForEventType retrieves enabled webhooks that should receive a specific event type.
+func (r *WebhooksRepository) ListEnabledForEventType(ctx context.Context, eventType string) ([]models.Webhook, error) {
+	query := `
```
BhagyaAmarasinghe, line 337: `ListEnabledForEventType` has no `ORDER BY` with a hard `LIMIT`. This query caps results at 1000 rows, but without an ORDER BY PostgreSQL returns rows in arbitrary order. If there are >1000 matching webhooks, which ones are included varies between invocations, meaning some webhooks may randomly miss events.
Add `ORDER BY id` (or `created_at`) before the `LIMIT` for deterministic behavior.

internal/workers/webhook_dispatch.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if !webhook.Enabled {
+		slog.Debug("webhook dispatch: webhook disabled, skipping",
+			"event_id", args.EventID,
+			"webhook_id", args.WebhookID,
+		)
+		return nil
+	}
+
+	payload := argsToPayload(args)
```
BhagyaAmarasinghe, line 62: Here also it looks like dead code. `argsToPayload` (line 113) always returns a non-nil pointer, it just constructs a struct literal and returns `&`. The `if payload == nil` check here can never be true. We can either remove this or add actual validation inside `argsToPayload` that could return `nil`.

internal/service/message_publisher.go
```diff
@@ -0,0 +1,112 @@
...
+func (m *MessagePublisherManager) PublishEventWithChangedFields(_ context.Context, eventType datatypes.EventType, data any, changedFields []string) {
+	event := Event{
+		ID:            uuid.Must(uuid.NewV7()),
+		Type:          eventType,
+		Timestamp:     time.Now().Unix(),
+		Data:          data,
+		ChangedFields: changedFields,
+	}
+
+	select {
```
BhagyaAmarasinghe, line 81: There's a silent event loss here. The `select/default` pattern silently drops events when the channel is full, with only a `slog.Warn`. This could cause missed deliveries with no operator visibility beyond logs.
We could at minimum expose a `metric/counter` for dropped events. A blocking send with a short timeout (`case m.eventChan <- event:` / `case <-time.After(...)`) could also be a better tradeoff than silent drops.

internal/service/webhooks_service.go
```diff
@@ -0,0 +1,151 @@
...
+		fields = append(fields, "tenant_id")
+	}
+	if req.EventTypes != nil {
+		fields = append(fields, "event_types")
+	}
+	return fields
+}
+
+// DeleteWebhook deletes a webhook by ID.
+func (s *WebhooksService) DeleteWebhook(ctx context.Context, id uuid.UUID) error {
```
BhagyaAmarasinghe, line 150: Here, `DeleteWebhook` has TOCTOU race (get-then-delete). Between `GetByID` (line 139) and `Delete` (line 144), another concurrent request could delete the same webhook, making `Delete` return `ErrNotFound`. We could use `DELETE ... RETURNING` (the same pattern used for `BulkDelete` in feedback records) to fetch and delete atomically, avoiding the extra round-trip and the race.

cmd/api/main.go
```diff
@@ -51,9 +56,44 @@ func run() int {
...
+		slog.Error("Failed to create River client", "error", err)
+		messageManager.Shutdown()
+		return exitFailure
+	}
+
+	webhookProvider := service.NewWebhookProvider(riverClient, webhooksRepo, cfg.WebhookDeliveryMaxAttempts, cfg.WebhookMaxFanOutPerEvent)
+	messageManager.RegisterProvider(webhookProvider)
+
+	// Start River in background
+	go func() {
```
BhagyaAmarasinghe: Here River start failure is silently dropped. If `riverClient.Start(ctx)` fails (e.g. missing River migrations), the error is only logged and the server keeps running without webhook delivery. There's no health check or error channel to detect this.
We could wait briefly for River to become healthy, propagating the error via a channel (similar to `serverErr`), or including River status in the `/health` endpoint.

internal/service/webhooks_service.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, fmt.Errorf("update webhook: %w", err)
+	}
+
+	s.publisher.PublishEventWithChangedFields(ctx, datatypes.WebhookUpdated, *webhook, s.getChangedFields(req))
+
+	return webhook, nil
+}
+
+// getChangedFields extracts which fields were changed from the update request.
+func (s *WebhooksService) getChangedFields(req *models.UpdateWebhookRequest) []string {
```
BhagyaAmarasinghe: `getChangedFields` methods don't use receiver. It never uses `s`. It should be a standalone function to make that clear:
`func getChangedFields(req *models.UpdateWebhookRequest) []string {`

internal/service/feedback_records_service.go
```diff
@@ -81,26 +88,70 @@ func (s *FeedbackRecordsService) UpdateFeedbackRecord(ctx context.Context, id uu
 	if err != nil {
 		return nil, fmt.Errorf("update feedback record: %w", err)
 	}
+
+	s.publisher.PublishEventWithChangedFields(ctx, datatypes.FeedbackRecordUpdated, *record, s.getChangedFields(req))
 	return record, nil
 }
 
+// getChangedFields extracts which fields were changed from the update request.
+func (s *FeedbackRecordsService) getChangedFields(req *models.UpdateFeedbackRecordRequest) []string {
```
BhagyaAmarasinghe: Here also `getChangedFields` doesn't use the `s` receiver. Convert to a standalone function.

internal/service/webhook_provider.go
```diff
@@ -0,0 +1,101 @@
...
+		)
+		return
+	}
+
+	if len(webhooks) == 0 {
+		return
+	}
+
+	// Cap fan-out per event to avoid blocking the publisher.
+	if len(webhooks) > p.maxFanOut {
```
BhagyaAmarasinghe: Here there's a random exclusion when fan-out cap kicks in. This caps to `maxFanOut` by slicing `webhooks[:p.maxFanOut]`. Combined with the lack of `ORDER BY` in `ListEnabledForEventType`), which webhooks get capped out is non-deterministic. After fixing the `ORDER BY`, consider also documenting or warning that older/newer webhooks get priority.

migrations/002_webhooks.sql
```diff
@@ -0,0 +1,20 @@
...
+  updated_at TIMESTAMP NOT NULL DEFAULT NOW()
+);
+
+CREATE INDEX idx_webhooks_enabled ON webhooks(enabled);
```
BhagyaAmarasinghe, line 23: Missing index for `ORDER BY.` The List query sorts by `created_at DESC` but there's no index on it. For small tables this is fine, but at scale this will cause a sequential scan + sort. 
We could add:
`CREATE INDEX idx_webhooks_created_at ON webhooks(created_at DESC);`

internal/api/handlers/webhooks_handler.go
```diff
@@ -0,0 +1,181 @@
...
+	service WebhooksService
+}
+
+// NewWebhooksHandler creates a new webhooks handler.
+func NewWebhooksHandler(service WebhooksService) *WebhooksHandler {
+	return &WebhooksHandler{service: service}
+}
+
+// Create handles POST /v1/webhooks.
+func (h *WebhooksHandler) Create(w http.ResponseWriter, r *http.Request) {
```
BhagyaAmarasinghe, line 38: Here there's no rate limit on webhook creation. A user (or compromised API key) could create thousands of webhooks. Since every event fans out to all enabled webhooks, this could create a massive amplification effect. Maybe we could add a max-webhooks-per-account guard in the service layer, or at minimum a creation rate limit?

internal/repository/webhooks_repository.go
```diff
@@ -0,0 +1,395 @@
...
+		req.URL, req.SigningKey, enabled, req.TenantID, eventTypes,
+	).Scan(
+		&webhook.ID, &webhook.URL, &webhook.SigningKey, &webhook.Enabled,
+		&webhook.TenantID, &webhook.CreatedAt, &webhook.UpdatedAt, &dbEventTypes,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create webhook: %w", err)
+	}
+
+	if dbEventTypes != nil {
```
BhagyaAmarasinghe: We can extract `dbEventTypes` parsing helper. This exact pattern (scan `[]string`, loop through `ParseEventType`, build `[]datatypes.EventType`) is repeated 5+ times in this file (lines 65-74, 101-109, 178-186, 308-316, 377-384). Extract a helper:
`func parseDBEventTypes(ss []string) ([]datatypes.EventType, error) { ... }`

internal/service/webhook_provider.go
```diff
@@ -0,0 +1,101 @@
...
+	"log/slog"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/riverqueue/river"
+	"github.com/riverqueue/river/rivertype"
+)
+
+// WebhookDispatchInserter inserts webhook_dispatch jobs in batch (e.g. River client).
+type WebhookDispatchInserter interface {
```
BhagyaAmarasinghe, line 14: compared to  [internal/service/message_publisher.go, line 18](https://github.com/formbricks/hub/pull/16/changes#diff-5c2d969b51493f6adad23766e672377ac011882e087f83103c37af86b46c65c5R18)
There's inconsistent timestamp types. `Event.Timestamp` is `int64` (Unix seconds) but `WebhookPayload.Timestamp` is `time.Time`. This forces a `time.Unix()` conversion in `argsToPayload`. Consider using the same type throughout, `time.Time` is generally preferred in Go and would avoid the lossy conversion.

tests/integration_test.go
```diff
@@ -560,3 +574,321 @@ func TestDeleteFeedbackRecord(t *testing.T) {
...
+	} {
+		body, err := json.Marshal(p)
+		require.NoError(t, err)
+		req, err := http.NewRequestWithContext(context.Background(), http.MethodPost, server.URL+"/v1/feedback-records", bytes.NewBuffer(body))
+		require.NoError(t, err)
+		req.Header.Set("Authorization", "Bearer "+testAPIKey)
+		req.Header.Set("Content-Type", "application/json")
+		resp, err := client.Do(req)
+		require.NoError(t, err)
+		defer func() { require.NoError(t, resp.Body.Close()) }()
```
BhagyaAmarasinghe: AI Nitpick:
In the create-records loop (lines ~596-608), defer func() { require.NoError(t, resp.Body.Close()) }() is called inside a for loop. These defers accumulate and only fire when the test function exits, not at the end of each iteration.
Not a high priority, good to merge even without addressing this.

tests/integration_test.go
```diff
@@ -560,3 +574,321 @@ func TestDeleteFeedbackRecord(t *testing.T) {
...
+	}
+
+	// Bulk delete by user_identifier
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodDelete, server.URL+"/v1/feedback-records?user_identifier="+userID, http.NoBody)
+	require.NoError(t, err)
+	req.Header.Set("Authorization", "Bearer "+testAPIKey)
+
+	resp, err := client.Do(req)
+	require.NoError(t, err)
+	defer func() { require.NoError(t, resp.Body.Close()) }()
```
BhagyaAmarasinghe: Here and following defers:
These defers accumulate and only fire when the test function exits, not at the end of each iteration. For a small loop this is fine, but it's a leak-prone pattern. Consider closing the body immediately after reading it:
```
read body...
resp.Body.Close()
```

BhagyaAmarasinghe (commented):
.

BhagyaAmarasinghe (changes requested):
Thanks for the PR :)
I have added several comments that needs to be looked.

internal/service/feedback_records_service.go
xernobyl: At some point I had a "feedback_record.bulk_delete" event, but that adds additional complexity on the events side.
I'm changing the "feedback_record.deleted" event to be sent with a list of deleted IDs, to make it more consistent, instead of sending the deleted object.

internal/api/handlers/webhooks_handler.go
xernobyl: Removed.

internal/workers/webhook_dispatch.go
xernobyl: I've noticed some of these before and similar ones, and I added some changes to the linter (another PR) to avoid this in the future.

internal/api/handlers/webhooks_handler.go
xernobyl: I was thinking of fine-tuning this kind of details later, but I'll put some hard limit for now...

internal/service/webhook_provider.go
xernobyl: That was a carry on from my previous attempt using a priority queue to handle the webhooks. Fixed.

internal/service/webhooks_service.go
xernobyl: Fixed.

internal/repository/webhooks_repository.go
xernobyl: I limited the number of webhooks, and I removed the limit. Limiting the number of webhooks should be good for now. I have some improvements to be done in the future, like caching the webhooks instead of looking up on the database every time there's a new event.

internal/service/message_publisher.go
xernobyl: The idea of this was for the event channel to be "big enough", and drops shouldn't really happen in practice... The only reason for events to over accumulate, is if the creation of jobs in river would be slow, but since the bottleneck there is the database writes, it's possible that would already fail when we're creating the resource the triggers the event.

The metrics are already planned in a future PR (already in draft).

I'll think a little bit more about this.

xernobyl: The buffer size is defined in an env var: MESSAGE_PUBLISHER_QUEUE_MAX_SIZE

internal/service/webhooks_service.go
xernobyl: GetByID is longer called, so this is fixed.

internal/service/webhooks_service.go
xernobyl: Fixed. Should be caught with the new linter rules (pending PR).

cmd/api/main.go
xernobyl: Refactored main. Adding river to health can be done later.

migrations/002_webhooks.sql
xernobyl: Done.

internal/repository/webhooks_repository.go
xernobyl: Done.

BhagyaAmarasinghe (approved):
Thank you, LGTM 🚀

PR #18, opened 2026-02-05
chore: make Postgres host port configurable via `POSTGRES_PORT`

xernobyl:
**Description:**

Makes the Postgres host port configurable so developers can avoid port conflicts (e.g. with the main Formbricks app) without changing shared config. Also updates the README for setup and environment variables.

**Configurable Postgres port**

- **compose.yml**: Port mapping uses `"${POSTGRES_PORT:-5432}:5432"` so the host port is controlled by env and defaults to 5432.
- **.env.example**: Documents optional `POSTGRES_PORT` and notes that `DATABASE_URL` must use the same port when overriding.
- **README**: Environment variables section is now a table (Variable, Required, Default, Description) including `POSTGRES_PORT` and its relation to `DATABASE_URL`. Troubleshooting added for port-already-in-use: set `POSTGRES_PORT` in `.env`, keep `DATABASE_URL` in sync, then `make docker-up`.
- **tests/README.md**: Prerequisites updated to describe `DATABASE_URL` and `POSTGRES_PORT` instead of the hardcoded port 5433.

**README: Go and PATH**

- **Prerequisites**: Go must be on `PATH`; added link to “Adding Go to PATH” and verification with `go version`.
- **Troubleshooting**: “Go not found” — install Go, follow Adding Go to PATH, restart terminal. “Docker/Postgres port in use” — see above (use `POSTGRES_PORT` and `DATABASE_URL`).
- **Adding Go to PATH**: New subsection with a single `export PATH="..."` line for macOS/Linux (bash/zsh), a note for the Go version manager, and verification with `which go`.

Default behaviour is unchanged (port 5432). To use another port, set `POSTGRES_PORT` and the same port in `DATABASE_URL` in your local `.env` only.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,10 +6,13 @@
 # The server will fail to start if this is not set
 API_KEY=your-secret-api-key-here
 
-# Database connection URL (optional)
-# Default: postgres://postgres:postgres@localhost:5433/test_db?sslmode=disable (compose exposes 5433 on host)
+# Postgres host port for docker-compose (optional). Default: 5432. Override to avoid conflicts (e.g. POSTGRES_PORT=5433); keep DATABASE_URL in sync.
+# POSTGRES_PORT=5432
+
+# Database connection URL (optional). Port must match POSTGRES_PORT when you override it.
+# Default: postgres://postgres:postgres@localhost:5432/test_db?sslmode=disable
 # Format: postgres://username:password@host:port/database?sslmode=disable
-DATABASE_URL=postgres://postgres:postgres@localhost:5433/test_db?sslmode=disable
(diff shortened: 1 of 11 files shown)


BhagyaAmarasinghe:
Right now the docs say `5432` but make run creates a `.env` with `5433` and doesn’t set P`OSTGRES_PORT`, so Compose stays on 5432 and things can get out of sync. Worth either aligning the Makefile with the docs (both 5432) or setting `POSTGRES_PORT=5433 `in the generated .env and adding a quick note that new setups use `5433` to avoid port clashes.

BhagyaAmarasinghe (approved):

PR #23, opened 2026-02-12
feat: hub dockerfile and helm chart

BhagyaAmarasinghe:
Hub dockerfile and helm chart


xernobyl (changes requested):
You have a few __pycache__ folders pushed. Is there a reason for those to be there?

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+# =============================================================================
+FROM golang:1.25.7-alpine AS builder
+
+RUN apk add --no-cache git ca-certificates
+
+WORKDIR /build
+
+# Install goose and river CLI for migrations
+RUN go install github.com/pressly/goose/v3/cmd/goose@v3.26.0 && \
+    go install github.com/riverqueue/river/cmd/river@latest
```
xernobyl: Pin that to the current version instead of latest.

xernobyl (changes requested):

Dockerfile
BhagyaAmarasinghe: fixed

xernobyl (approved):
LGTM

PR #28, opened 2026-02-13
feat: Observability (metrics, tracing, request correlation)

xernobyl:
### Observability
- **Metrics**: Prometheus at `GET /metrics` when `OTEL_METRICS_EXPORTER=prometheus`. Hub metrics: event pipeline (discarded events, channel depth, fan-out duration), webhook delivery (enqueued, deliveries by status, duration histograms, disabled counts), River queue depth. HTTP metrics via otelhttp.
- **Tracing**: Optional OTLP or stdout when `OTEL_TRACES_EXPORTER` is set. HTTP spans via otelhttp; trace context (`trace_id`, `span_id`) and `request_id` added to structured logs when using `*Context` slog methods.
- **Request correlation**: `X-Request-ID` middleware (propagate or generate UUID v7, echo in response).
- **Docs**: New [Metrics and Observability](/reference/metrics), env var reference for `OTEL_*`, suggested alerts and practices.

Metrics and tracing are opt-in; if env vars are unset, a startup warning is logged and `/metrics` is not registered.

### Other
- README and auth docs: removed references to `todo.md` for OpenAPI.


cmd/api/app.go
```diff
@@ -12,32 +12,113 @@ import (
...
+	meterProvider  *sdkmetric.MeterProvider
+	tracerProvider *sdktrace.TracerProvider
+	eventMetrics   observability.EventMetrics
+}
+
+const riverQueueDepthInterval = 15 * time.Second
+
+// setupMetrics creates meter provider, handler, event metrics, and webhook metrics when metrics are enabled.
+func setupMetrics(cfg *config.Config) (
+	*sdkmetric.MeterProvider, http.Handler, observability.EventMetrics, observability.WebhookMetrics, error,
```
BhagyaAmarasinghe: There's a crash bug here where nil-pointer panic when `OTEL_METRICS_EXPORTER` is set to anything other than "prometheus"
In app.go, if someone sets `OTEL_METRICS_EXPORTER=otlp` (a valid OTel value, just not supported here). `NewMeterProvider` returns `(nil, nil, nil)` for non-`"prometheus"` values (no error), then `mp.Meter("hub")` panics on nil. 
We need to either return an error from `NewMeterProvider` for unsupported values, or add a nil check on `mp` in `setupMetrics`.

internal/observability/tracing.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
+	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+
+	"github.com/formbricks/hub/internal/config"
+)
+
+func newOTLPTraceExporter(ctx context.Context, cfg *config.Config) (sdktrace.SpanExporter, error) {
```
BhagyaAmarasinghe: Here, OTLP endpoint handling bypasses SDK env var parsing and mishandles URL vs host:port.
`otlptracehttp.WithEndpoint()` expects a bare `host:port` (e.g., localhost:4318), but the documentation and env var convention use full URLs (http://localhost:4318). Also, the Go OTel SDK natively reads `OTEL_EXPORTER_OTLP_ENDPOINT` and correctly parses scheme/host/port/insecure. By manually reading it into config and passing it via `WithEndpoint()`, you bypass the SDK's built-in handling and introduce a mismatch. This will cause real connection failures with SigNoz setups.
We could either;
1. Remove the manual config field and `WithEndpoint()` call entirely. Let the SDK read `OTEL_EXPORTER_OTLP_ENDPOINT` natively. Just call `otlptracehttp.New(ctx)`. - Simplest approach.
2. If you need config control: Parse the URL, extract host:port, and conditionally add `otlptracehttp.WithInsecure()` for http:// schemes.

cmd/api/app.go
```diff
@@ -12,32 +12,113 @@ import (
...
+	} else {
+		meterProvider, metricsHandler, eventMetrics, webhookMetrics, err = setupMetrics(cfg)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	var tracerProvider *sdktrace.TracerProvider
+
+	if cfg.OtelTracesExporter == "" {
```
BhagyaAmarasinghe, line 104: `TraceContextHandler` (and request_id in logs) is only installed when tracing is enabled. The `RequestID` middleware always runs, but the `TraceContextHandler` that actually adds `request_id` to log records is only installed when tracing is enabled. If you enable metrics but not tracing (a valid config), `request_id` never appears in logs. The handler wrapping should be unconditional (it already handles missing spans gracefully).

internal/observability/provider.go
```diff
@@ -0,0 +1,135 @@
...
+	if err := provider.Shutdown(ctx); err != nil {
+		return fmt.Errorf("meter provider shutdown: %w", err)
+	}
+
+	return nil
+}
+
+// NewTracerProvider creates a TracerProvider when tracing is enabled.
+// When cfg.OtelTracesExporter is empty, returns (nil, nil).
+func NewTracerProvider(cfg *config.Config) (*sdktrace.TracerProvider, error) {
```
BhagyaAmarasinghe, line 90: This defaults to `AlwaysOnSampler`, which in production with high traffic will generate excessive trace data and can measurably impact performance. We can support configurable sampling (`OTEL_TRACES_SAMPLER` with `always_on`, `always_off`, `traceidratio`, `parentbased_traceidratio`).

cmd/api/app.go
```diff
@@ -103,6 +211,20 @@ func newHTTPServer(
...
+	otelOpts := []otelhttp.Option{}
+	if meterProvider != nil {
+		otelOpts = append(otelOpts, otelhttp.WithMeterProvider(meterProvider))
+	}
+
+	if tracerProvider != nil {
+		otelOpts = append(otelOpts, otelhttp.WithTracerProvider(tracerProvider))
+	}
+
+	// Handler chain: built bottom-to-top; request flows RequestID -> Logging -> otelhttp -> mux.
```
BhagyaAmarasinghe: Here, `trace_id`/`span_id` won't appear in HTTP access logs.
The middleware chain is `RequestID -> Logging -> otelhttp -> mux`. The logging middleware captures r`.Context()` before `otelhttp` adds the span to context. When the logging middleware calls `slog.InfoContext(r.Context(), "HTTP request", ...) `after `next.ServeHTTP` returns, it's using the context from the RequestID middleware, not the otelhttp-enriched context. So `trace_id`/`span_id` will be absent from access logs. To fix this, the logging middleware would need to run inside the otelhttp handler, or the response-writer wrapper needs to capture the enriched context.

cmd/api/app.go
```diff
@@ -103,6 +211,20 @@ func newHTTPServer(
 	mux.Handle("/v1/", protectedWithAuth)
 	mux.Handle("/", public)
 
+	otelOpts := []otelhttp.Option{}
```
BhagyaAmarasinghe: `/health` and `/metrics` generate traces and HTTP metrics (noise). The `otelhttp.NewHandler` wraps everything including `/health` and `/metrics`.
We could add something like this:
```
otelOpts = append(otelOpts, otelhttp.WithFilter(func(r *http.Request) bool {
    return r.URL.Path != "/health" && r.URL.Path != "/metrics"
}))
```

internal/service/webhook_provider.go
```diff
@@ -39,6 +47,10 @@ func NewWebhookProvider(inserter WebhookDispatchInserter, repo WebhooksRepositor
 func (p *WebhookProvider) PublishEvent(ctx context.Context, event Event) {
```
BhagyaAmarasinghe, line 47: `RecordJobsEnqueued` is inaccurate on partial batch failure
The method returns early on InsertMany error, so if batch 3 of 4 fails, the metric records 0 total instead of the actual count from batches 1-2. Track enqueued count per-batch for accuracy:
```
var enqueued int64
// inside loop after successful InsertMany:
enqueued += int64(len(batch))
// after loop:
if p.metrics != nil {
    p.metrics.RecordJobsEnqueued(event.Type.String(), enqueued)
}
```

cmd/api/app.go
```diff
@@ -153,23 +279,97 @@ func (a *App) Run(ctx context.Context) error {
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			update()
+		}
+	}
+}
+
 // Shutdown stops the server, message publisher, and River in order. Call after Run returns.
```
BhagyaAmarasinghe: The `Shutdown` method repeats the same tracer/meter provider shutdown sequence 3 times. A `defer`-based approach or a helper would be cleaner:
```
func (a *App) Shutdown(ctx context.Context) error {
    defer a.message.Shutdown()
    defer func() {
        if a.tracerProvider != nil {
            if err := observability.ShutdownTracerProvider(ctx, a.tracerProvider); err != nil {
                slog.Error("shutdown tracer provider", "error", err)
            }
        }
        if a.meterProvider != nil {
            if err := observability.ShutdownMeterProvider(ctx, a.meterProvider); err != nil {
                slog.Error("shutdown meter provider", "error", err)
            }
        }
    }()
    // ... rest of shutdown
}
```

internal/observability/names.go
```diff
@@ -0,0 +1,83 @@
...
+	AttrStatus    = "status"
+)
+
+// AllowedEventTypes returns event type strings allowed for metric attributes (bounded cardinality).
+func AllowedEventTypes() []string {
+	return datatypes.GetAllEventTypes()
+}
+
+// AllowedProviderReasons for hub_webhook_provider_errors_total.
+var AllowedProviderReasons = map[string]bool{
```
BhagyaAmarasinghe: Lines 35 to 56: `AllowedProviderReasons`, `AllowedDeliveryStatuses`, etc are exported `var map[string]bool`. Any importing package could mutate them at runtime. Either make them unexported or use functions that return copies.

internal/observability/provider.go
```diff
@@ -0,0 +1,135 @@
...
+	reg := prometheus.NewRegistry()
+
+	exporter, err := prometheusexporter.New(
+		prometheusexporter.WithRegisterer(reg),
+	)
+	if err != nil {
+		return nil, nil, fmt.Errorf("create prometheus exporter: %w", err)
+	}
+
+	res, err := resource.Merge(
```
BhagyaAmarasinghe, line 19: Lines 35-44 and 86-95 are identical. Extract into a shared helper.

cmd/api/app.go
```diff
@@ -64,27 +161,38 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
 	feedbackRecordsHandler := handlers.NewFeedbackRecordsHandler(feedbackRecordsService)
 	healthHandler := handlers.NewHealthHandler()
 
-	server := newHTTPServer(cfg, healthHandler, feedbackRecordsHandler, webhooksHandler)
+	server := newHTTPServer(cfg, healthHandler, feedbackRecordsHandler, webhooksHandler, meterProvider, tracerProvider, metricsHandler)
```
BhagyaAmarasinghe, line 179: Providers are stored on the struct and passed to `otelhttp` explicitly, but `otel.SetTracerProvider()` / `otel.SetMeterProvider()` are never called. Any library or future code using `otel.GetTracerProvider()` won't pick them up. Consider registering globally as a convenience.

internal/observability/events.go
```diff
@@ -0,0 +1,113 @@
...
+	evtMetrics.riverQueueGauge = riverQueueGauge
+
+	return evtMetrics, nil
+}
+
+func attrEventType(v string) attribute.KeyValue {
+	return attribute.String(AttrEventType, v)
+}
+
+func (e *eventMetrics) RecordEventDiscarded(eventType string) {
```
BhagyaAmarasinghe: Using `context.Background()` in every `Add()`/`Record()` call means no trace context is associated with metric data points. For future exemplar support (linking a metric sample to a specific trace), the actual request/worker context should be passed through the metrics interfaces. Should accept `ctx context.Context` as a parameter for future exemplar support.
same repeats in:
https://github.com/formbricks/hub/pull/28/changes#diff-fc794729e1171e38c3e231c938229960217033e403fcd5dd837d538e9008c8dfR102
and
https://github.com/formbricks/hub/pull/28/changes#diff-0b0384d7a1533ac627fe499da7ee4e2cb29315c537568e9395c736a2b7720954R98

BhagyaAmarasinghe (changes requested):
Thanks for the PR :)
 I have mentioned several issues I noticed and also the current implementation uses Prometheus pull-based metrics (`/metrics` endpoint) while using OTLP push for traces. Since we're standardizing on OTel + SigNoz, we should use OTLP push for both signals.
This reduces dependencies as it drops `prometheus/client_golang`, `promhttp`, and the `OTel` Prometheus exporter.

Replace `prometheusexporter` + `promhttp` with `otlpmetrichttp` + `PeriodicExportingMetricReader` (e.g., 60s export interval, matching ee). This also removes the `/metrics` endpoint and the Prometheus registry.

cmd/api/app.go
xernobyl: Fixed.

cmd/api/app.go
xernobyl: Now it's always installed. Fixed.

internal/observability/provider.go
xernobyl: Added sampler env var.

internal/observability/tracing.go
xernobyl: Fixed.

cmd/api/app.go
xernobyl: Global registration is in place now.

internal/observability/events.go
xernobyl: Added context param.

internal/observability/provider.go
xernobyl: Done.

internal/observability/names.go
xernobyl: Fixed.

cmd/api/app.go
xernobyl: Fixed.

cmd/api/app.go
xernobyl: Fixed.

internal/service/webhook_provider.go
xernobyl: Fixed.

cmd/api/app.go
xernobyl: Fixed.

BhagyaAmarasinghe (approved):

PR #31, opened 2026-02-16
feat: ci and release workflow

BhagyaAmarasinghe:
CI and release workflow

### DIFF
--- a/.github/workflows/build-and-push-ecr.yml
+++ b/.github/workflows/build-and-push-ecr.yml
@@ -1,0 +1,162 @@
+name: Build & Push to AWS ECR
+
+# Builds Formbricks Hub Docker images for ECR deployment:
+# - workflow_call: Used by release workflow with explicit SemVer versions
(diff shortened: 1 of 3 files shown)


mattinannt (approved):
@BhagyaAmarasinghe thanks a lot for the PR :-) 

Looks good!
I'm wondering if building the image two times (once for ECR and once for GCR) is something we can combine and one build and two separate pushes, but that's an optimization we can do later when we are certain about the use cases and we know when we build the images (e.g. only on release).

BhagyaAmarasinghe:
> @BhagyaAmarasinghe thanks a lot for the PR :-)
> 
> Looks good! I'm wondering if building the image two times (once for ECR and once for GCR) is something we can combine and one build and two separate pushes, but that's an optimization we can do later when we are certain about the use cases and we know when we build the images (e.g. only on release).

I added as 2 for now until we test out the release workflow and iron out all the issues. Once its done, we can combine into a single build.

PR #32, opened 2026-02-17
fix: build for linux/arm64 so image runs on ARM Macs

xernobyl:
Use TARGETOS/TARGETARCH in Dockerfile so multi-platform builds (CI already uses linux/amd64,linux/arm64) produce correct binaries for each architecture. Fixes exit 127 when running hub image on Apple Silicon.


BhagyaAmarasinghe (approved):
Thank you :)

PR #34, opened 2026-02-18
chore: adds river to install tools

pandeymangg:
adds river to `install-tools` as well in the Makefile


xernobyl (approved):
Thank you for the PR! Looks good.

PR #38, opened 2026-02-23
feat: AI embeddings for feedback records

xernobyl:
## What does this PR do?

Adds **embeddings for feedback records**: when a feedback record is created or updated and has non-empty `value_text`, the system enqueues a job to generate an embedding via a configurable provider (OpenAI or Google Gemini) and stores it in a dedicated `embeddings` table (pgvector). This keeps embedding data out of the main feedback-records read path and supports multiple models per record.

**Highlights:**
- **Event-driven**: `EmbeddingProvider` subscribes to `feedback_record.created` and `feedback_record.updated` (when `value_text` is in changed fields). Jobs are enqueued to a dedicated River queue (`embeddings`) so embedding work does not starve webhook delivery.
- **Pluggable providers**: Single event provider and worker; embedding API is behind an `EmbeddingClient` interface. Implementations: **OpenAI** (`text-embedding-3-small`, configurable dimensions) and **Google Gemini** (`gemini-embedding-001`, via `google.golang.org/genai`). API key is optional (e.g. for local AI).
- **Schema**: New `embeddings` table: `id`, `feedback_record_id`, `embedding` (vector), `model`, `created_at`, `updated_at`; unique on `(feedback_record_id, model)`; `ON DELETE CASCADE` from `feedback_records`.
- **Backfill**: `cmd/backfill-embeddings` enqueues jobs for existing records that have `value_text` but no embedding for the configured model. Requires `EMBEDDING_PROVIDER` and `EMBEDDING_MODEL` (no defaults).
- **Config**: `EMBEDDING_PROVIDER` (openai | google), `EMBEDDING_MODEL`, `EMBEDDING_PROVIDER_API_KEY` (optional), `EMBEDDING_DIMENSIONS` (default 1536), `EMBEDDING_MAX_CONCURRENT`, `EMBEDDING_MAX_ATTEMPTS`. Supported providers are kept in a map for easy extension.
- **Observability**: Embedding metrics (enqueued, outcomes, duration, errors) and structured logging; worker retries on transient errors and skips retry on not-found.

Fixes #(issue)

No API contract changes: embeddings are not exposed on feedback-record list/get. They are stored for future use (e.g. semantic search).

## How should this be tested?

- **Unit**: `go test ./internal/... ./cmd/...` (includes `embedding_provider_test.go`, worker tests).
- **Integration**: `make tests` (requires Postgres; `tests/` use feedback records service with embedding model; DB must have `embeddings` table from migration).
- **Manual**:
  1. `make init-db` (and `make river-migrate` if using River UI) so `embeddings` exists.
  2. Set in `.env`: `EMBEDDING_PROVIDER=openai`, `EMBEDDING_MODEL=text-embedding-3-small`, `EMBEDDING_PROVIDER_API_KEY=sk-...` (or use `google` and a Gemini API key; or leave key empty for a no-key provider).
  3. `make run`; create a feedback record with `value_text`; confirm embedding job is enqueued and processed (logs: "embedding: job enqueued", "embedding: stored") and a row appears in `embeddings`.
  4. Backfill: `EMBEDDING_PROVIDER=openai EMBEDDING_MODEL=text-embedding-3-small DATABASE_URL=... go run ./cmd/backfill-embeddings` (both env vars required).

### Required

- [x] Ran `make build`
- [ ] Merged the latest changes from main onto my branch with `git pull origin main`

### Appreciated

- [ ] If API changed: added or updated OpenAPI spec and ran contract tests (`make tests` or API contract workflow)
- [ ] If API behavior changed: added request/response examples or Swagger UI screenshots to this PR


internal/service/embedding_provider.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// On create, only enqueue when there is embeddable text. On update we enqueue regardless so the worker can clear.
+	if event.Type == datatypes.FeedbackRecordCreated &&
+		(record.ValueText == nil || strings.TrimSpace(*record.ValueText) == "") {
+		slog.Debug("embedding: skip, no value_text on create", "feedback_record_id", record.ID)
+
+		return
+	}
+
+	opts := &river.InsertOpts{
```
mattinannt, line 87: AI Feedback:

Embedding jobs are deduplicated only by feedback_record_id (ByArgs: true) with a 24-hour uniqueness window. That means a record edited twice within 24h can have its second enqueue dropped even though value_text changed, leaving a stale embedding until the window expires. The ticket explicitly calls for idempotency keyed by input hash/provider/model; current dedupe semantics are too coarse for updates.

internal/workers/feedback_embedding.go
```diff
@@ -0,0 +1,188 @@
...
+func (w *FeedbackEmbeddingWorker) Timeout(*river.Job[service.FeedbackEmbeddingArgs]) time.Duration {
+	return feedbackEmbeddingTimeout
+}
+
+// Work loads the record, generates or clears the embedding, and persists it.
+func (w *FeedbackEmbeddingWorker) Work(ctx context.Context, job *river.Job[service.FeedbackEmbeddingArgs]) error {
+	args := job.Args
+	start := time.Now()
+
+	record, err := w.embeddingService.GetFeedbackRecord(ctx, args.FeedbackRecordID)
```
mattinannt, line 62: AI Feedback:

All GetFeedbackRecord errors are treated as final and the worker returns nil, but the comment only mentions the not-found case. This also swallows transient DB/network failures and permanently loses jobs that should be retried. The worker should distinguish not-found from retryable errors before suppressing retries.

internal/repository/feedback_records_repository.go
```diff
@@ -76,19 +113,22 @@ func (r *FeedbackRecordsRepository) GetByID(ctx context.Context, id uuid.UUID) (
 			source_type, source_id, source_name,
 			field_id, field_label, field_type, field_group_id, field_group_label,
 			value_text, value_number, value_boolean, value_date,
-			metadata, language, user_identifier, tenant_id
+			metadata, language, user_identifier, tenant_id, embedding
```
mattinannt: AI Feedback:

GetByID, List, and Update now select and scan the embedding vector from feedback_records for normal API reads. Combined with the new embedding field in the API model/OpenAPI, this makes all feedback record responses much heavier and exposes internal vectors on endpoints unrelated to search. That directly conflicts with the ticket's 'API performance unaffected' and 'lean schema' goals.

mattinannt (changes requested):
@xernobyl Thank you for the PR :-) 

Please always add the ticket link to the PR description so that ticket and PR get linked.

I currently only did a small review with a first AI-based review. Also maybe we need to discuss how we would proceed with supporting multiple AI/embeddings providers as this might happen much sooner than we might think.

internal/repository/feedback_records_repository.go
xernobyl: I wasn't sure if it was needed outside of queries. Hiding it for now.

internal/service/embedding_provider.go
xernobyl: Added an hash to distinguish them.

internal/workers/feedback_embedding.go
xernobyl: Fixed.

cmd/api/app.go
```diff
@@ -136,10 +159,53 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
 	riverWorkers := river.NewWorkers()
```
BhagyaAmarasinghe, line 164: `FeedbackRecordsService` is constructed twice, its first built with `nil` as the River client, then re-built after the River client is created. 
The `embeddingWorker` captures a reference to the first service instance (with nil inserter). It works today because the worker only calls `GetFeedbackRecord` and `SetEmbedding`, but this is a maintenance trap. If anyone later adds code to the worker that calls `BackfillEmbeddings`, it will nil-pointer panic.
We can use a setter method or lazy initialization instead.

internal/workers/feedback_embedding.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	err = w.embeddingService.SetEmbedding(ctx, args.FeedbackRecordID, args.Model, embedding)
+	if err != nil {
+		if w.metrics != nil {
+			w.metrics.RecordWorkerError(ctx, "update_failed")
+			w.metrics.RecordEmbeddingOutcome(ctx, "failed_final")
+			w.metrics.RecordEmbeddingDuration(ctx, time.Since(start), "failed_final")
+		}
+
```
BhagyaAmarasinghe, line 127: Here worker silently swallows final-attempt failures. When the embedding API fails on the final attempt, the worker returns nil, so River marks the job as successfully completed. These records will never get embeddings and won't show as failed jobs in River UI or monitoring. 
We can return the error to let River mark it as discarded/failed, or at least record a metric specifically for "permanently failed" embeddings so they're discoverable.

internal/openai/client.go
```diff
@@ -0,0 +1,113 @@
...
+	if len(resp.Data) == 0 {
+		return nil, ErrNoEmbeddingInResponse
+	}
+
+	emb := resp.Data[0].Embedding
+	if len(emb) != c.dimensions {
+		return nil, fmt.Errorf("%w: got %d, want %d", ErrDimensionMismatch, len(emb), c.dimensions)
+	}
+
+	out := make([]float32, len(emb))
```
BhagyaAmarasinghe, line 104: here openai SDK returns float64, converted to float32 with precision loss. This is fine for embeddings (precision beyond float32 doesn't matter for cosine similarity), but worth a brief comment noting the intentional truncation.

cmd/backfill-embeddings/main.go
```diff
@@ -0,0 +1,133 @@
...
+	exitSuccess                 = 0
+	exitFailure                 = 1
+)
+
+func main() {
+	os.Exit(run())
+}
+
+func run() int {
+	databaseURL := os.Getenv("DATABASE_URL")
```
BhagyaAmarasinghe, line 45: The main API server loads .env via godotenv.Load(), but the backfill command reads env vars directly from os.Getenv. Add .env loading for consistency.

migrations/004_add_feedback_records_embedding.sql
```diff
@@ -0,0 +1,28 @@
...
+  embedding vector NOT NULL,
+  model TEXT NOT NULL,
+  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  UNIQUE (feedback_record_id, model)
+);
+
+-- HNSW index for cosine similarity search (OpenAI text-embedding-3-small).
+-- NOTICE the double parentheses and the explicit ::vector(1536) cast
+CREATE INDEX idx_embeddings_openai ON embeddings USING hnsw ((embedding::vector(1536)) vector_cosine_ops)
```
BhagyaAmarasinghe: I'm wondering here at lines 17-23, If a user configures a different model (eg text-embedding-3-large at 3072 dims), no HNSW index will cover their queries. This should at minimum be documented, or the migration should be parameterized / a separate migration step added for custom models.

internal/config/config.go
```diff
@@ -142,6 +158,21 @@ func Load() (*Config, error) {
 		return nil, ErrWebhookMaxCount
 	}
 
+	embeddingDimensions := getEnvAsInt("OPENAI_EMBEDDING_DIMENSIONS", defaultEmbeddingDimensions)
```
BhagyaAmarasinghe: Since this supports Google Gemini too, the env var should be renamed to` EMBEDDING_DIMENSIONS` for consistency with the other `EMBEDDING_*` vars.

BhagyaAmarasinghe (changes requested):
Thanks for the PR, I have commented on some issues I've noticed.
Could you also check on below 2 points as well:

1. in feedback_records.go `CreateFeedbackRecordRequest.SubmissionID` is a non-pointer string with validate:"required". Every existing API consumer that doesn't send submission_id will now get a 400 validation error. This is a breaking change that should be called out in a changelog or migration guide, or made optional with a server-generated default.

2. Migration 003 adds NOT NULL column with no default. If the table has any existing rows, this ALTER TABLE will fail because PostgreSQL cannot add a NOT NULL column without a DEFAULT value to a table with existing data. The migration needs a strategy: add as nullable, backfill (e.g set submission_id = id::text), then add the NOT NULL constraint.

internal/config/config.go
xernobyl: Forgot to rename that one... Fixed.

migrations/004_add_feedback_records_embedding.sql
xernobyl: Documented. We need a different index for each model and embedding size.

cmd/backfill-embeddings/main.go
xernobyl: Fixed.

internal/openai/client.go
xernobyl: Commented, and changed embeddings to 16 bits :D

internal/workers/feedback_embedding.go
xernobyl: Added logging.

cmd/api/app.go
xernobyl: Fixed.

xernobyl:
@BhagyaAmarasinghe submission_id topics can be fixed on another PR

BhagyaAmarasinghe (approved):

migrations/004_add_feedback_records_embedding.sql
```diff
@@ -0,0 +1,29 @@
...
+
+-- HNSW index for cosine similarity search (OpenAI text-embedding-3-small, 1536 dimensions).
+-- halfvec_cosine_ops: same semantics as vector_cosine_ops for half-precision column.
+-- Other models (e.g. text-embedding-3-large at 3072 dims) are not covered; add a separate index if needed.
+CREATE INDEX idx_embeddings_openai ON embeddings USING hnsw ((embedding::halfvec(1536)) halfvec_cosine_ops)
+WHERE model = 'text-embedding-3-small';
+
+-- HNSW index for cosine similarity search (Google gemini-embedding-001, 768 dimensions).
+CREATE INDEX idx_embeddings_google ON embeddings USING hnsw ((embedding::halfvec(768)) halfvec_cosine_ops)
+WHERE model = 'gemini-embedding-001';
```
mattinannt: The models all have variable vector sizes that you can configure in the API call, so you can also use the Gemini embedding model, for example, with 512-dimensional vectors. 

Also, new models launch regularly, the users might use a version of the hub which doesn't know that the  `gemini-embedding-002` model has already been released. Then they cannot use the model, which is not ideal. 

So I would definitely avoid mentioning direct model names in the code. We should only think about what different provider authentication mechanisms exist, and then the dimension as well as the model name can freely be set by the user. What options do we have to achieve this?

cmd/api/app.go
```diff
@@ -136,10 +159,53 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
...
 	river.AddWorker(riverWorkers, webhookWorker)
 
+	queues := map[string]river.QueueConfig{
+		river.QueueDefault:          {MaxWorkers: cfg.WebhookDeliveryMaxConcurrent},
+		service.EmbeddingsQueueName: {MaxWorkers: cfg.EmbeddingMaxConcurrent},
+	}
+
+	feedbackRecordsRepo := repository.NewFeedbackRecordsRepository(db)
+	embeddingsRepo := repository.NewEmbeddingsRepository(db)
+	embeddingProviderName, embeddingModel := embeddingProviderAndModel(cfg)
```
mattinannt, line 174: AI Feedback:

**[P1] Embeddings are enabled by default even when no provider API key is configured**

NewApp registers the embedding worker/provider whenever the provider name is supported, but it does not require EmbeddingProviderAPIKey. Since config defaults to EMBEDDING_PROVIDER=openai and EMBEDDING_MODEL=text-embedding-3-small, a fresh deployment will start enqueueing embedding jobs and the worker will call the provider with an empty API key, causing repeated job failures/noise. This also contradicts the config comment that embeddings are enabled only when the API key is set."

---

We should not have defaults for the embedding provider. If no embedding provider is set, then the user doesn't get embeddings, and also the embedding jobs should not run. I think for now for the hub embeddings should be optional.

mattinannt (changes requested):
@xernobyl thank you for updating the PR. I have a few concerns regarding the embedding model provider and indexes. let's discuss.

cmd/api/app.go
xernobyl: Defaulting to none.

migrations/004_add_feedback_records_embedding.sql
xernobyl: In that case you can set the embedding dimension and model in the env vars. The only thing that is missing is the creation of the index, and that doesn't prevent the Hub from working, just makes the operation less performant. Can be mentioned in the documentation.
```

EMBEDDING_PROVIDER=openai
EMBEDDING_PROVIDER_API_KEY=xxx
EMBEDDING_MODEL=text-embedding-3-small
EMBEDDING_DIMENSIONS=1536

#EMBEDDING_PROVIDER=google
#EMBEDDING_PROVIDER_API_KEY=xxx
#EMBEDDING_MODEL=gemini-embedding-001
#EMBEDDING_DIMENSIONS=768
```

That can be any model those APIs support.
We can programmatically create an index for the selected settings.

mattinannt (approved):

PR #39, opened 2026-02-23
feat: add submission_id for grouped submissions and idempotent ingestion

xernobyl:
## What does this PR do?

Adds a `submission_id` field to `feedback_records` so that records belonging to one logical feedback submission can be grouped. This supports use cases like multi-field ingestion from a single event (e.g. Stripe cancellation: reason, comment, date) and simplifies:

- **Grouped reads**: `GET /v1/feedback-records?submission_id=...`
- **Idempotent webhook retries**: duplicate (tenant_id, submission_id, field_id) returns 409 Conflict
- **Partial-write reconciliation** and correlation across ingestion pipelines

**Changes:**
- **Migration** `003_add_feedback_records_submission_id.sql`: adds nullable `submission_id` (VARCHAR 255), indexes for filtering, and partial unique index on `(tenant_id, submission_id, field_id)` where `submission_id IS NOT NULL` (one value per field per submission; backward compatible for rows without submission_id).
- **Models**: `FeedbackRecord` and create/list request types include `submission_id`; list filters support `submission_id`.
- **Repository**: Create/Get/List/Update read/write `submission_id`; Create maps PostgreSQL unique violation (23505) to `huberrors.ConflictError`.
- **Handler**: Create responds with 409 Conflict and message when duplicate (tenant_id, submission_id, field_id).
- **OpenAPI**: Create body and list params document `submission_id`; FeedbackRecordData includes `submission_id`; 409 response documented for create.
- **Tests**: `TestFeedbackRecordsSubmissionID` (create with submission_id, list by submission_id), `TestFeedbackRecordsSubmissionIDUniqueConstraint` (duplicate create returns 409).

Existing integrations remain compatible: `submission_id` is optional on create; records without it are valid and the unique constraint does not apply to NULL.

**Example – create with submission_id:**
```json
POST /v1/feedback-records
{
  "source_type": "formbricks",
  "submission_id": "550e8400-e29b-41d4-a716-446655440000",
  "tenant_id": "org-123",
  "field_id": "reason",
  "field_type": "text",
  "value_text": "cancelled"
}
```

**Example – list by submission_id:**
```
GET /v1/feedback-records?submission_id=550e8400-e29b-41d4-a716-446655440000&tenant_id=org-123
```

**Example – duplicate create (same tenant_id, submission_id, field_id):** returns `409 Conflict` with body describing the conflict.

## How should this be tested?

- **Unit tests**: `make test-unit` (passes).
- **Integration tests**: Start Postgres (`make docker-up`), then `make tests`. Covers create without/with submission_id, list by submission_id, and duplicate create → 409.
- **Build/lint**: `make build`, `make fmt`, `make lint` (all pass).
- **Migrations**: `make init-db` to apply migration; `make migrate-validate` to validate migration files.
- **Manual**: Create two records with same `tenant_id`, `submission_id`, and `field_id`; second request should return 409. List with `?submission_id=...` returns only records with that submission_id.

### Required

- [x] Ran `make build`
- [x] Ran `make tests` (integration tests require Postgres; unit tests and build pass)

### Appreciated

Solves https://github.com/formbricks/internal/issues/1424

### DIFF
--- a/migrations/003_add_feedback_records_submission_id.sql
+++ b/migrations/003_add_feedback_records_submission_id.sql
@@ -1,20 +1,19 @@
 -- +goose up
 -- Add submission_id to feedback_records for grouping records belonging to one logical submission.
 -- Enables idempotent multi-field ingestion and simpler grouped reads (e.g. GET ?submission_id=...).
+-- Mandatory (NOT NULL): every record must belong to a submission; use e.g. field_id if single-field.
 
 ALTER TABLE feedback_records
-  ADD COLUMN submission_id VARCHAR(255);
+  ADD COLUMN submission_id VARCHAR(255) NOT NULL;
 
 -- Index for filtering by submission_id (and tenant-scoped list by submission)
 CREATE INDEX idx_feedback_records_submission_id ON feedback_records(submission_id);
 CREATE INDEX idx_feedback_records_tenant_submission_id ON feedback_records(tenant_id, submission_id);
 
 -- One value per field_id per submission per tenant (supports idempotent webhook retries).
--- Partial index: only enforce uniqueness when submission_id is set (backward compatible for NULL).
 -- NULLS NOT DISTINCT (PG 15+): treat NULL tenant_id as equal so duplicate (NULL, submission_id, field_id) conflicts.
 CREATE UNIQUE INDEX idx_feedback_records_tenant_submission_field
-  ON feedback_records(tenant_id, submission_id, field_id) NULLS NOT DISTINCT
-  WHERE submission_id IS NOT NULL;
+  ON feedback_records(tenant_id, submission_id, field_id) NULLS NOT DISTINCT;
 
 -- +goose down
 DROP INDEX IF EXISTS idx_feedback_records_tenant_submission_field;
--- a/migrations/003_add_feedback_records_submission_id.sql
+++ b/migrations/003_add_feedback_records_submission_id.sql
@@ -11,8 +11,9 @@ CREATE INDEX idx_feedback_records_tenant_submission_id ON feedback_records(tenan
 
 -- One value per field_id per submission per tenant (supports idempotent webhook retries).
 -- Partial index: only enforce uniqueness when submission_id is set (backward compatible for NULL).
+-- NULLS NOT DISTINCT (PG 15+): treat NULL tenant_id as equal so duplicate (NULL, submission_id, field_id) conflicts.
 CREATE UNIQUE INDEX idx_feedback_records_tenant_submission_field
-  ON feedback_records(tenant_id, submission_id, field_id)
+  ON feedback_records(tenant_id, submission_id, field_id) NULLS NOT DISTINCT
   WHERE submission_id IS NOT NULL;
 
 -- +goose down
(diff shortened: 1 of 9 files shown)


migrations/003_add_feedback_records_submission_id.sql
```diff
@@ -0,0 +1,22 @@
...
+
+ALTER TABLE feedback_records
+  ADD COLUMN submission_id VARCHAR(255);
+
+-- Index for filtering by submission_id (and tenant-scoped list by submission)
+CREATE INDEX idx_feedback_records_submission_id ON feedback_records(submission_id);
+CREATE INDEX idx_feedback_records_tenant_submission_id ON feedback_records(tenant_id, submission_id);
+
+-- One value per field_id per submission per tenant (supports idempotent webhook retries).
+-- Partial index: only enforce uniqueness when submission_id is set (backward compatible for NULL).
```
mattinannt: Since we haven't launched the Hub yet, I would propose that we make the `submission_id` mandatory and don't allow NULL values to avoid edge cases.

migrations/003_add_feedback_records_submission_id.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE UNIQUE INDEX idx_feedback_records_tenant_submission_field
+  ON feedback_records(tenant_id, submission_id, field_id)
+  WHERE submission_id IS NOT NULL;
```
mattinannt: AI Feedback:

The new uniqueness guarantee is implemented as a partial unique index on (tenant_id, submission_id, field_id), but tenant_id is still nullable in the schema and remains optional in the create API. In PostgreSQL, UNIQUE does not consider rows with NULL values equal, so duplicate creates with the same submission_id and field_id but no tenant_id will not conflict. That weakens the intended idempotent retry behavior for a currently valid request shape.

-----

I'm also unsure how we should solve this. What are your ideas?

mattinannt (changes requested):
@xernobyl Thanks a lot for the PR :-) Already looks very good.

Only two points I think we should address before we can merge this:

migrations/003_add_feedback_records_submission_id.sql
xernobyl: I've addressed this by making the unique index use NULLS NOT DISTINCT, with that, NULLs in the unique columns are treated as equal, so duplicate creates that share submission_id and field_id and omit tenant_id get a unique violation.

migrations/003_add_feedback_records_submission_id.sql
xernobyl: If it's mandatory should we have a default value? Otherwise, the users will need to come up with one.

mattinannt: we don't have a default value. it's mandatory that the user needs to set this. But of course this assumes that a feedbackRecord always belongs to a submission. But if that's not the case they can also set it to the same value as the `field_id`.

But of course the approach of having this mandatory doesn't work if it's already in use and needs a migration. But since we haven't launched, I'm not aware of any system already using the Hub.

xernobyl: Now mandatory.

mattinannt (approved):

PR #40, opened 2026-02-25
feat: semantic search and similar feedback API

xernobyl:
Adds **semantic search** and **similar feedback** endpoints, and improves embedding configuration and robustness.

### Semantic search & similar feedback

- **POST /v1/feedback-records/search/semantic** – embeds the query and returns feedback records with similarity scores (cosine, 0..1)
- **GET /v1/feedback-records/{id}/similar** – returns feedback records similar to a given one by embedding

Both endpoints require embeddings to be configured (EMBEDDING_PROVIDER + EMBEDDING_MODEL). When disabled, they return 503 Service Unavailable. Features:

- Query embedding cache (LRU) for repeated searches
- Keyset (cursor-based) pagination for stable, efficient paging
- `hasMore` flag so pagination works correctly when the repo over-fetches and filters by `minScore`

### Embedding configuration changes

- **EMBEDDING_PROVIDER** and **EMBEDDING_MODEL** both required to enable embeddings; no default model
- **EMBEDDING_PROVIDER_API_KEY** required for openai and google; server fails fast if missing
- Provider name normalized to lowercase
- **EMBEDDING_NORMALIZE** env var for L2-normalizing vectors before storage
- `EmbeddingPrefixForProvider` future proofing support for other embedding providers

### Backfill-embeddings

- Creates embedding client directly (openai/google) with same options as API
- Uses `FeedbackEmbeddingWorker` for job processing
- Requires `.env`; fails fast if missing
- `make run-backfill-embeddings` target

### Other changes

- **JSON decode errors**: `response.JSONDecodeErrorDetail` for clearer 400 messages on invalid request bodies
- **OpenAPI**: realistic examples for parameters; `maximum: 1000` for search offset; search endpoints documented
- **CI**: Schemathesis split – main step excludes search (503); separate step tests search endpoints with relaxed checks
- **Migration 005**: `CREATE INDEX CONCURRENTLY` / `DROP INDEX CONCURRENTLY` for `idx_embeddings_model` (no write blocking)
- **Normalize**: precompute `invMag` and multiply instead of divide in L2 normalization loop

### Unit tests

```bash
make test-unit
```

### Integration tests

```bash
make dev-setup   # or: docker-up, init-db
make tests
```

### Semantic search (requires embeddings)

1. Configure `.env`:
   ```
   EMBEDDING_PROVIDER=openai
   EMBEDDING_MODEL=text-embedding-3-small
   EMBEDDING_PROVIDER_API_KEY=sk-...
   ```
2. Start API: `make run`
3. Backfill embeddings: `make run-backfill-embeddings`
4. Semantic search:
   ```bash
   curl -X POST "http://localhost:8080/v1/feedback-records/search/semantic" \
     -H "Authorization: Bearer $API_KEY" \
     -H "Content-Type: application/json" \
     -d '{"query":"login is slow","tenant_id":"org-123"}'
   ```

### API contract tests

```bash
make schemathesis   # with API running; runs search endpoints when embeddings enabled
```

### Required

- [x] Read [Repository Guidelines](AGENTS.md)
- [x] Ran `make build`
- [x] Ran `make tests` (integration tests)
- [x] Merged the latest changes from main onto my branch
- [x] If database schema changed: migration 005 uses CONCURRENTLY; ran `make migrate-validate`

### Appreciated

- [x] If API changed: OpenAPI spec updated; contract tests in CI
- [x] Updated docs if necessary


internal/service/embedding_job_args.go
mattinannt, line 22: To increase the accuracy of the embedding, I think it makes sense to embed not only the `value_text` value but rather `field_label + value_text` to add more information to the embedding.

e.g.
`field_label: What features are you missing in Formbricks?`
`value_text: Dashboards, Charts`

I'm not sure about the best way to concatenate the strings so AI can read them best (e.g. just append; key value structure, text formatting with newlines,...), but maybe you can check on any recommendations for this. Embeddings are similar to LLMs, so maybe a bit more context about what those fields are might be helpful to the embeddings models.

This would also make the `field_label` an important information we should return from the similarity endpoints.

openapi.yaml
```diff
@@ -544,6 +544,139 @@ paths:
                                             - location: "body.value_text"
                                               message: "Text value cannot be empty"
                                               value: ""
+    /v1/feedback-records/search/semantic:
```
mattinannt, line 567: I don't see the `tenant_id` here as part of the request / params. We need to be able to only get feedbackRecords from a specific tenant.

openapi.yaml
```diff
@@ -544,6 +544,139 @@ paths:
...
+                  in: path
+                  description: Feedback record ID (UUID)
+                  required: true
+                  schema:
+                    type: string
+                    format: uuid
+                - name: tenantId
+                  in: query
+                  description: Tenant ID (required for isolation; must match feedback record tenant_id)
+                  required: true
```
mattinannt, line 669: Currently in the Hub datamodel the `tenant_id` is optional.
I'm thinking if that decision was a mistake that makes things more complicated along the way since it created more edge cases.

Maybe we should just make the tenant_id mandatory in the data model before we launch. What do you think?

cmd/backfill-embeddings/main.go
```diff
@@ -10,20 +10,32 @@ import (
...
 	"github.com/formbricks/hub/internal/repository"
 	"github.com/formbricks/hub/internal/service"
+	"github.com/formbricks/hub/internal/workers"
 	"github.com/formbricks/hub/pkg/database"
 )
 
+const (
+	embeddingProviderOpenAI = "openai"
+	embeddingProviderGoogle = "google"
+	defaultEmbeddingModelDB = "default"
```
mattinannt: What is the `default` model?

I think there shouldn't be something like this.  Either the environment variable is set and then we can enable the embeddings workers + the similarity search endpoint; if it's not set, embeddings generation should not be enabled and the similarity search endpoint should be disabled.

internal/service/search_service.go
```diff
@@ -0,0 +1,225 @@
...
+// If cursor is non-empty it is used for keyset paging (offset is ignored); otherwise offset is used.
+func (s *SearchService) SimilarFeedback(
+	ctx context.Context, feedbackRecordID uuid.UUID, tenantID string, limit, offset int, minScore float64, cursor string,
+) (SearchResult, error) {
+	out := SearchResult{}
+	if tenantID == "" {
+		return out, ErrMissingTenantID
+	}
+
+	embedding, err := s.embeddingsRepo.GetEmbeddingByFeedbackRecordAndModel(ctx, feedbackRecordID, s.model)
```
mattinannt: AI Feedback:

**similar can use a source embedding from another tenant**

The source vector is loaded by feedback_record_id + model only, without checking tenantId. Neighbor results are tenant-filtered later, but the query vector itself can come from a different tenant’s record.
Practical impact: this violates the documented contract (“tenant must match source record”) and weakens tenant isolation semantics.

internal/repository/embeddings_repository.go
```diff
@@ -90,3 +102,173 @@ func (r *EmbeddingsRepository) ListFeedbackRecordIDsForBackfill(ctx context.Cont
...
+		err  error
+	)
+
+	if excludeID == nil {
+		rows, err = r.db.Query(ctx, `
+			SELECT e.feedback_record_id, (1 - (e.embedding <=> $1)) AS score, fr.value_text
+			FROM embeddings e
+			INNER JOIN feedback_records fr ON fr.id = e.feedback_record_id
+			WHERE e.model = $2 AND fr.tenant_id = $3 AND (1 - (e.embedding <=> $1)) >= $4
+			ORDER BY e.embedding <=> $1
```
mattinannt: AI Feedback:

**Pagination order differs between offset and cursor modes**

Offset query orders by distance only; cursor query orders by (distance, feedback_record_id).
Practical impact: when multiple rows share the same distance, switching from page 1 (offset) to page 2 (cursor) can skip or duplicate rows because tie-breaking changed mid-stream.

openapi.yaml
mattinannt, line 567: AI Feedback:

**Routes are conditionally missing although OpenAPI documents them unconditionally**

If embeddings are disabled, search handlers are not registered, so clients get generic 404. Docs still advertise endpoints.
Practical impact: clients cannot distinguish “feature disabled” from “wrong URL,” and generated SDK behavior won’t match runtime expectations.


----

I think it's correct behavior that the endpoints are not working, if embeddings are not enabled. But we should disable the endpoints with e.g. `503 Service Unavailable` and a message, not just omit them.

mattinannt (commented):
@xernobyl Thanks for the PR :-) 

Looking good already; we just need some fine tuning :-) 

I'll also test the functionality in more detail in the coming days and also build a small use-case to test this with real data.

internal/repository/embeddings_repository.go
```diff
@@ -90,3 +102,173 @@ func (r *EmbeddingsRepository) ListFeedbackRecordIDsForBackfill(ctx context.Cont
...
+) ([]models.FeedbackRecordWithScore, error) {
+	if len(queryEmbedding) != models.EmbeddingVectorDimensions {
+		return nil, fmt.Errorf("%w: got %d, want %d", ErrEmbeddingDimensionMismatch, len(queryEmbedding), models.EmbeddingVectorDimensions)
+	}
+
+	if offset < 0 {
+		offset = 0
+	}
+
+	queryVec := pgvector.NewHalfVector(queryEmbedding)
```
BhagyaAmarasinghe: The query vector is being converted to `HalfVector` here (`pgvector.NewHalfVector(queryEmbedding)`), which quantizes it down to float16. The stored embeddings use halfvec to save space, and that's a reasonable trade-off. But there's no reason to also lose precision on the query side, the query is ephemeral and doesn't get stored. pgvector can compare a full-precision `vector` against a `halfvec` column natively, so this should use `pgvector.NewVector(queryEmbedding)` instead. The same applies to line 216 in `NearestFeedbackRecordsByEmbeddingAfterCursor`. Keeping the query at float32 eliminates one source of quantization noise from every search.

internal/repository/embeddings_repository.go
```diff
@@ -90,3 +102,173 @@ func (r *EmbeddingsRepository) ListFeedbackRecordIDsForBackfill(ctx context.Cont
...
+
+	var (
+		rows pgx.Rows
+		err  error
+	)
+
+	if excludeID == nil {
+		rows, err = r.db.Query(ctx, `
```
BhagyaAmarasinghe: I see two issues with this query. 
First, `hnsw.ef_search` is never set before executing the search. pgvector defaults to `ef_search = 40`, which means the HNSW index only considers 40 candidate nodes during graph traversal. For any non-trivial dataset this significantly limits recall. A `SET LOCAL hnsw.ef_search = 200` (or a configurable value) should be issued within the same transaction before this query runs. This is the single highest-impact change for search quality and costs essentially nothing in terms of implementation effort.

Second, the `WHERE` clause includes `(1 - (e.embedding <=> $1)) >= $4` as a minimum-score filter. This prevents pgvector's iterative index scan from working. Iterative scan is what allows pgvector to automatically fetch more candidates from the HNSW index when rows are filtered out by other conditions (like `tenant_id` or `model`). With the distance threshold in the `WHERE` clause, Postgres treats the entire query as a filtered scan and won't iterate back into the index for more candidates. The min-score threshold should be moved to application-level filtering, fetch results ordered by distance with a limit, then discard any below the threshold in Go code. The same issue applies to the cursor-based query starting at line 225.

Also, the `INNER JOIN feedback_records fr ... WHERE fr.tenant_id = $3` is a post-filter on HNSW results. If only a small fraction of embeddings belong to the queried tenant, the index returns its `ef_search `candidates and most get discarded by the tenant filter, leaving fewer results than requested. Fixing the iterative scan issue (by removing the distance threshold from WHERE) will partially address this, since pgvector will then automatically pull more candidates. But for deployments with many tenants and skewed data distributions, consider over-fetching internally (eg request `3 * limit` from the index, filter, then trim to `limit`) as an additional safety net.

migrations/004_add_feedback_records_embedding.sql
BhagyaAmarasinghe, line 16: The HNSW index is created with default parameters (`m = 16, ef_construction = 64`). These defaults optimize for fast index builds, not for search recall. For a production similarity search feature, `m = 24` and `ef_construction = 200 `are more appropriate. Higher `m` means each node in the HNSW graph has more connections (better recall, slightly more storage), and higher `ef_construction` means the index builder considers more candidates when inserting nodes (better graph quality, slower builds). This is a one-time cost at index build time that improves every subsequent query. The index should be: `CREATE INDEX idx_embeddings ON embeddings USING hnsw (embedding halfvec_cosine_ops) WITH (m = 24, ef_construction = 200)`.

internal/googleai/client.go
BhagyaAmarasinghe: The `EmbedContentConfig` is only setting `OutputDimensionality` but not `TaskType`. Google's embedding models are trained to produce different vector representations depending on whether the input is a document being stored or a query being searched. By passing `TaskType: "RETRIEVAL_DOCUMENT"` when embedding feedback records and `TaskType: "RETRIEVAL_QUERY" `when embedding search queries, the model shifts the vectors into a space that's optimized for asymmetric retrieval. This is a free quality improvement that requires no reindexing or infrastructure changes, just two different code paths for document vs. query embedding. The `CreateEmbedding` method would need a parameter (or a separate method) to distinguish between the two use cases.

BhagyaAmarasinghe (dismissed):
Thank you for the PR and I've been checking how we could improve the similarity search and found out a couple of areas to improve where I've commented. Can you also do a quick search on the below 2 general suggestions which I believe will also help a lot to improve the similarity search:

1. Sentence-level chunking for multi-topic feedback
Feedback records often contain multiple topics in a single response, for example, "The UI is great and I love the new features. But the app crashes every time I hit save, it's useless." When this entire paragraph is embedded as a single vector, the result is a semantic average of positive-UI and negative-stability signals. That averaged vector won't strongly match a query about either topic. Chunking feedback at the sentence level (or using a sliding window with overlap) and storing each chunk as its own embedding would solve this: the sentence about crashing would have a vector that cleanly aligns with bug-related queries without contamination from the positive UI comment. This is a significant architectural change, it requires moving from a 1:1 relationship between feedback records and embeddings to a 1:N relationship, a new table or schema, and modified search queries that deduplicate results by parent record. It's worth considering if feedback records in practice tend to be long and multi-topic, but may not be necessary if most records are short, single-topic responses.

2. Using HyDE (Hypothetical Document Embeddings) as an optional search mode
There is a fundamental semantic gap between how people phrase search queries and how feedback is actually written. A query like "what are people saying about slow login" looks nothing like the feedback it should match: "I waited 5 minutes just to sign in, this is ridiculous." Standard embedding models encode these two strings into very different regions of vector space because their structure, tone, and vocabulary are completely different. HyDE addresses this by passing the search query through an LLM first, asking it to generate a hypothetical feedback response that would answer the query. That synthetic response is then embedded and used as the search vector instead of the raw query. Because a fake review and a real review share sentence structure and vocabulary, cosine similarity between them is dramatically higher. The trade-off is an additional LLM round-trip per search request (typically 200-800ms latency plus token cost), so this should be offered as an optional mode for complex or analytical queries rather than the default path for every search.

internal/service/embedding_job_args.go
xernobyl: Content of the generated embeddings was changed to include question and answer pairs, along with the return results.

openapi.yaml
xernobyl: Makes sense. tenant_id changed to mandatory, further improvements will be done in the future.

cmd/backfill-embeddings/main.go
xernobyl: That was accidentally left there. There's no default, default is disabled.

internal/service/search_service.go
xernobyl: Fixed: The source vector for “similar” is only loaded when the record belongs to the requested tenant. The earlier concern about using another tenant’s embedding is addressed.

internal/repository/embeddings_repository.go
xernobyl: Fixed.

openapi.yaml
xernobyl: Returning 503 when service is disabled.

internal/repository/embeddings_repository.go
xernobyl: Fixed.

internal/repository/embeddings_repository.go
xernobyl: Issues have been addressed.

internal/googleai/client.go
xernobyl: Implemented that.

migrations/004_add_feedback_records_embedding.sql
xernobyl: Fixed.

openapi.yaml
xernobyl: Fixed.

mattinannt (commented):
AI Review by Code

1) [P3] internal/config/config.go (around line 59)
AI Review comment by Code: `EMBEDDING_MODEL` is documented as optional in config comments / `.env.example`, but runtime behavior now effectively requires it to enable embeddings (and backfill explicitly hard-fails without it). This creates operator confusion because the documented contract does not match actual behavior.
How to resolve: update config comments and `.env.example` to state that embeddings require both `EMBEDDING_PROVIDER` and `EMBEDDING_MODEL` (and clarify provider-specific API-key requirements).

2) [P2] internal/repository/embeddings_repository.go (around line 195)
AI Review comment by Code: in both nearest-neighbor methods, a deferred `Rollback` logs any error unconditionally. After a successful `Commit`, `Rollback` returns `pgx.ErrTxClosed`, so successful requests can emit false error logs. This can create noisy and misleading production telemetry.
How to resolve: in the deferred rollback block, ignore `pgx.ErrTxClosed` (or track commit success and skip rollback logging after commit).

3) [P2] cmd/api/app.go (around line 178)
AI Review comment by Code: the API app enables embeddings when provider+model are set, but does not validate required API key presence for providers that need one (OpenAI/Google). Backfill now validates this, so app/backfill behavior is inconsistent. In app mode this can surface as runtime 500s from embedding calls instead of a clear startup/config failure.
How to resolve: add provider-specific API-key validation in app startup (same policy as backfill), then either fail fast with a clear startup error or disable embeddings consistently with an explicit log message.

internal/config/config.go
```diff
@@ -184,6 +205,7 @@ func Load() (*Config, error) {
 		EmbeddingModel:          getEnv("EMBEDDING_MODEL", ""),
```
mattinannt, line 205: AI Review comment by Code: Runtime behavior now requires `EMBEDDING_MODEL` (together with `EMBEDDING_PROVIDER`) to enable embeddings, and backfill also fails without it. But config/docs still describe the model as optional. This mismatch can mislead operators and cause confusing setup/debugging. Suggested fix: update config comments and `.env.example` to state the effective enablement rule clearly, including provider-specific API-key requirements.

internal/repository/embeddings_repository.go
```diff
@@ -90,3 +113,246 @@ func (r *EmbeddingsRepository) ListFeedbackRecordIDsForBackfill(ctx context.Cont
...
+
+	fetchLimit := min(limit*nearestOverFetchFactor+offset, maxNearestFetchLimit)
+
+	dbTx, err := r.db.BeginTx(ctx, pgx.TxOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("begin tx: %w", err)
+	}
+
+	defer func() {
+		if err := dbTx.Rollback(ctx); err != nil {
```
mattinannt: AI Review comment by Code: This deferred rollback logs any rollback error unconditionally. After a successful `Commit`, `Rollback` returns `pgx.ErrTxClosed`, so successful requests can emit false error logs. The same pattern appears in the cursor method below. Suggested fix: ignore `pgx.ErrTxClosed` in rollback logging (or track commit success and skip rollback handling post-commit).

cmd/api/app.go
```diff
@@ -188,17 +195,21 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
...
 	)
 
+	var searchHandler *handlers.SearchHandler
+
 	if embeddingProviderName != "" {
 		var embeddingClient service.EmbeddingClient
 
 		switch embeddingProviderName {
 		case embeddingProviderOpenAI:
 			embeddingClient = openai.NewClient(cfg.EmbeddingProviderAPIKey,
```
mattinannt, line 214: AI Review comment by Code: App startup enables embeddings/search when provider+model are set, but it does not validate API key presence for providers that require one (OpenAI/Google). Backfill now validates this, so behavior is inconsistent across entrypoints. In API mode this can degrade into runtime 500s instead of a clear startup/config error. Suggested fix: apply the same provider-specific API-key validation in app startup and either fail fast with a clear message or disable embeddings consistently with explicit logging.

mattinannt (changes requested):
AI Review by Code

Requesting changes for the inline points below.

mattinannt:
@coderabbitai

internal/config/config.go
xernobyl: Check wasn't being done because specific providers may not need API keys (like local instances). Changed error message and check to be more specific.

internal/repository/embeddings_repository.go
xernobyl: Added additional check for errtxclosed.

cmd/api/app.go
xernobyl: Fixed.

mattinannt:
@coderabbitai

mattinannt:
@coderabbitai please review

openapi.yaml
mattinannt, line 1: Following the default OpenAPI documentation I faced the following error in the `POST /v1/feedback-records` route:

```
{
    "type": "about:blank",
    "title": "Validation Error",
    "status": 400,
    "detail": "validation failed: SubmissionID is required"
}
```


Please make sure all feedback records have the `submission_id` marked as mandatory.

openapi.yaml
```diff
@@ -762,6 +911,61 @@ components:
...
+                - tenantId
+        SemanticSearchResponse:
+            type: object
+            additionalProperties: false
+            properties:
+                results:
+                    type: array
+                    description: List of feedback record IDs with similarity scores (0 = unrelated, 1 = identical)
+                    items:
+                        $ref: '#/components/schemas/SemanticSearchResultItem'
```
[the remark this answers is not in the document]
xernobyl: Done.

migrations/005_embeddings_indexes_and_feedback_tenant_required.sql
```diff
@@ -0,0 +1,19 @@
...
+-- B-tree index on model to help filter embeddings by model before/while using the HNSW vector index.
+CREATE INDEX idx_embeddings_model ON embeddings (model);
+
+-- Make tenant_id required for feedback_records. Backfill existing NULLs with empty string for legacy data.
+UPDATE feedback_records SET tenant_id = '' WHERE tenant_id IS NULL;
+ALTER TABLE feedback_records ALTER COLUMN tenant_id SET NOT NULL;
+
+-- Recreate HNSW index with m=24, ef_construction=200 for better recall (defaults are m=16, ef_construction=64).
+DROP INDEX IF EXISTS idx_embeddings;
+CREATE INDEX idx_embeddings ON embeddings USING hnsw (embedding halfvec_cosine_ops) WITH (m = 24, ef_construction = 200);
```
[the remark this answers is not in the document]
xernobyl: Done.

migrations/005_embeddings_indexes_and_feedback_tenant_required.sql
```diff
@@ -0,0 +1,19 @@
+-- +goose Up
+-- B-tree index on model to help filter embeddings by model before/while using the HNSW vector index.
+CREATE INDEX idx_embeddings_model ON embeddings (model);
+
+-- Make tenant_id required for feedback_records. Backfill existing NULLs with empty string for legacy data.
+UPDATE feedback_records SET tenant_id = '' WHERE tenant_id IS NULL;
+ALTER TABLE feedback_records ALTER COLUMN tenant_id SET NOT NULL;
```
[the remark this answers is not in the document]
xernobyl: Done.

internal/service/search_service_test.go
```diff
@@ -0,0 +1,229 @@
...
+	if m.createFunc != nil {
+		return m.createFunc(ctx, input)
+	}
+
+	return []float32{0.1}, nil
+}
+
+func (m *mockEmbeddingClient) CreateEmbeddingForQuery(ctx context.Context, input string) ([]float32, error) {
+	return m.CreateEmbedding(ctx, input)
+}
```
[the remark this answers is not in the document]
xernobyl: Done.

internal/service/search_cursor.go
```diff
@@ -0,0 +1,65 @@
...
+// EncodeSearchCursor returns an opaque cursor for the next page. distance is the cosine distance
+// (e.embedding <=> query) of the last result row; id is that row's feedback_record_id.
+func EncodeSearchCursor(distance float64, id uuid.UUID) string {
+	b, err := json.Marshal(cursorPayload{D: distance, I: id.String()})
+	if err != nil {
+		return ""
+	}
+
+	return base64.URLEncoding.EncodeToString(b)
+}
```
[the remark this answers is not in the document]
xernobyl: Done.

internal/service/feedback_records_service.go
```diff
@@ -52,6 +52,7 @@ type FeedbackRecordsService struct {
 }
 
 // NewFeedbackRecordsService creates a new feedback records service.
+// publisher may be nil when the service is used only for backfill (BackfillEmbeddings does not use the publisher).
 // embeddingInserter and embeddingQueueName are optional (for backfill); when nil/empty, BackfillEmbeddings returns an error.
 // Call SetEmbeddingInserter after the River client is created to enable backfill without building the service twice.
 // embeddingsRepo and embeddingModel are required for SetEmbedding and BackfillEmbeddings (from EMBEDDING_PROVIDER and EMBEDDING_MODEL).
```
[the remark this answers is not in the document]
xernobyl: Done.

openapi.yaml
```diff
@@ -762,6 +911,61 @@ components:
...
+                query:
+                    type: string
+                    description: Search query text (embedded and compared via cosine similarity)
+                topK:
+                    type: integer
+                    description: Maximum number of results (default 10, max 100)
+                    default: 10
+                    minimum: 1
+                    maximum: 100
+                tenantId:
```
mattinannt: In Postman I see the `tenant_id` just with value `string` here.

Image

For the feedback record create endpoint I had it being set to `org-123` by default. I think for the user it would be the easiest way if we already fill default values correctly and consistent across the schema and thus also here have it already been set to `org-123`.

Image

openapi.yaml
```diff
@@ -762,6 +911,61 @@ components:
             required:
                 - deleted_count
                 - message
+        SemanticSearchInputBody:
```
mattinannt, line 961: Sending the request in Postman without any custom change to the values gives me this error. We should make sure to have requests ideally working without any changes for testing. Not sure if we can deselect fields / make them optional.

Image

Image

openapi.yaml
```diff
@@ -762,6 +911,61 @@ components:
...
+            properties:
+                feedbackRecordId:
+                    type: string
+                    format: uuid
+                    description: Feedback record UUID
+                score:
+                    type: number
+                    format: double
+                    description: Similarity score (0..1)
+                fieldLabel:
```
mattinannt: Why is this camel case?

I think we should return the same names as we have in the database. So `field_label` for the label as well as `value_text` for the text (instead of `text`)

openapi.yaml
```diff
@@ -762,6 +911,61 @@ components:
...
+                    type: string
+                    description: Label of the feedback field (included in embedding for context)
+                value:
+                    type: string
+                    description: value_text of the feedback record (the text that was embedded); omitted if null
+                    nullable: true
+            required:
+                - feedbackRecordId
```
mattinannt: same here, this should also be snake_case for consistency.

openapi.yaml
```diff
@@ -544,6 +544,155 @@ paths:
...
+                  schema:
+                    type: integer
+                    minimum: 0
+                    default: 0
+                - name: cursor
+                  in: query
+                  description: Opaque cursor from previous response nextCursor for keyset paging. When set, offset is ignored and the next page after the cursor is returned (better performance for deep paging).
+                  schema:
+                    type: string
+                - name: minScore
```
mattinannt: We have a mix of snake case and lower camel case. To make it easy for developers we should stick with one casing.

let's use `snake_case` to be consistent with the data model. So please also change `minScore` to `min_score`

mattinannt: Can we set a reasonable default for the min_score to give good results without much noise? e.g. a `0.7`

openapi.yaml
```diff
@@ -762,6 +911,61 @@ components:
...
                 - deleted_count
                 - message
+        SemanticSearchInputBody:
+            type: object
+            additionalProperties: false
+            properties:
+                query:
+                    type: string
+                    description: Search query text (embedded and compared via cosine similarity)
+                topK:
```
mattinannt: In my testing topK below 4 threw internal server error:

Image

```
time=2026-03-03T17:45:36.428+01:00 level=ERROR msg="nearest feedback records: rollback failed" error="tx is closed"
time=2026-03-03T17:45:36.428+01:00 level=ERROR msg="semantic search: nearest failed" error="commit: conn busy" model=gemini-embedding-001
time=2026-03-03T17:45:36.428+01:00 level=INFO msg="HTTP request" method=POST path=/v1/feedback-records/search/semantic status=500 duration=3.303625ms request_id=019cb497-3d28-7ba0-9b58-1cf47d95bd1c
time=2026-03-03T17:45:37.559+01:00 level=ERROR msg="nearest feedback records: commit failed" error="conn busy"
time=2026-03-03T17:45:37.559+01:00 level=ERROR msg="nearest feedback records: rollback failed" error="tx is closed"
time=2026-03-03T17:45:37.559+01:00 level=ERROR msg="semantic search: nearest failed" error="commit: conn busy" model=gemini-embedding-001
time=2026-03-03T17:45:37.559+01:00 level=INFO msg="HTTP request" method=POST path=/v1/feedback-records/search/semantic status=500 duration=6.950333ms request_id=019cb497-4190-7c38-9e03-68b257d41182
time=2026-03-03T17:45:40.419+01:00 level=ERROR msg="nearest feedback records: rollback failed" error="tx is closed"
time=2026-03-03T17:45:40.419+01:00 level=INFO msg="HTTP request" method=POST path=/v1/feedback-records/search/semantic status=200 duration=7.095833ms request_id=019cb497-4cbc-7484-9802-99382fc036c1
time=2026-03-03T17:46:24.444+01:00 level=ERROR msg="nearest feedback records: commit failed" error="conn busy"
time=2026-03-03T17:46:24.445+01:00 level=ERROR msg="nearest feedback records: rollback failed" error="tx is closed"
time=2026-03-03T17:46:24.445+01:00 level=ERROR msg="semantic search: nearest failed" error="commit: conn busy" model=gemini-embedding-001
time=2026-03-03T17:46:24.445+01:00 level=INFO msg="HTTP request" method=POST path=/v1/feedback-records/search/semantic status=500 duration=6.993917ms request_id=019cb497-f8b6-7090-9fb7-9057cd5c61c8
```

mattinannt (changes requested):
@xernobyl Thank you for the updates to the PR.
While testing I noticed a few issues with the endpoints. The most pressing thing for me is the inconsistency in naming switching between lower camel case and snake case.
We should be consistent to reduce the mental load of developers and prevent bugs due to this.

Most findings I had regarding the semantic-search endpoint are also relevant for the similar feedbackRecord endpoint, so please also check for this.

openapi.yaml
xernobyl: Fixed.

openapi.yaml
xernobyl: Fixed.

openapi.yaml
xernobyl: Fixed.

openapi.yaml
xernobyl: Fixed.

openapi.yaml
xernobyl: Fixed.

xernobyl: Fixed.

openapi.yaml
xernobyl: Openapi.yml improved / error message improved.

mattinannt:
@coderabbitai please review

mattinannt (approved):
Thanks for the changes; works great! :-)

.env.example
```diff
@@ -43,12 +43,15 @@ WEBHOOK_MAX_FAN_OUT_PER_EVENT=500
...
+# For openai and google, EMBEDDING_PROVIDER_API_KEY is required; other providers (e.g. local) may not need it.
 # Vector size is fixed (768) in code.
 # EMBEDDING_PROVIDER=openai
-# EMBEDDING_PROVIDER_API_KEY=sk-...   (optional for some providers, e.g. local)
-# EMBEDDING_MODEL=<model name>        (optional; provider may have its own default)
+# EMBEDDING_PROVIDER_API_KEY=sk-...   (required for openai and google)
+# EMBEDDING_MODEL=<model name>        (required to enable embeddings; no default)
+
+# Search rate limit: max requests per minute per API key for semantic search and similar-feedback endpoints. 0 = disabled. Default: 60.
+# SEARCH_RATE_LIMIT_REQUESTS_PER_MINUTE=60
```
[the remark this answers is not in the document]
xernobyl: Was deprecated.

internal/api/handlers/search_handler.go
```diff
@@ -0,0 +1,283 @@
...
+	if err != nil {
+		return 0
+	}
+
+	if val < 0 {
+		return 0
+	}
+
+	return math.Min(val, 1)
+}
```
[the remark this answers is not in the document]
xernobyl: Returning default value now.

internal/api/response/response.go
```diff
@@ -46,6 +50,53 @@ func RespondBadRequest(w http.ResponseWriter, detail string) {
...
+
+	var buf strings.Builder
+
+	for i, r := range fieldPath {
+		if i > 0 && unicode.IsUpper(r) && (i+1 >= len(fieldPath) || unicode.IsLower(rune(fieldPath[i+1]))) {
+			buf.WriteByte('_')
+		}
+
+		buf.WriteRune(unicode.ToLower(r))
+	}
```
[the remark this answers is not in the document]
xernobyl: Using library now.

internal/models/feedback_records.go
```diff
@@ -101,8 +101,8 @@ type FeedbackRecord struct {
 	Metadata        json.RawMessage `json:"metadata,omitempty"`
 	Language        *string         `json:"language,omitempty"`
 	UserIdentifier  *string         `json:"user_identifier,omitempty"`
-	TenantID        *string         `json:"tenant_id,omitempty"`
-	SubmissionID    *string         `json:"submission_id,omitempty"`
+	TenantID        string          `json:"tenant_id"`
+	SubmissionID    string          `json:"submission_id"` // mandatory; never null (DB NULL is exposed as "")
```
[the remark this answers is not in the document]
xernobyl: Fixed.

.github/workflows/api-contract-tests.yml
```diff
@@ -119,9 +119,13 @@ jobs:
           checks: not_a_server_error,status_code_conformance,content_type_conformance,response_schema_conformance
           # Only run examples phase for fast, deterministic CI testing.
           # For deeper testing (stateful, fuzzing), run locally: make schemathesis
+          # Exclude search endpoints: CI does not configure embeddings, so they return 503 (treated as failure by not_a_server_error).
+          # Run locally with embeddings enabled to test these operations.
           args: >-
             --phases examples
             -H "Authorization: Bearer test-api-key-for-ci"
+            --exclude-operation-id semantic-search-feedback-records
+            --exclude-operation-id similar-feedback-records
```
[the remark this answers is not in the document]
xernobyl: Fixed.

Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: help tests tests-coverage build build-backfill-embeddings run init-db clean docker-up docker-down docker-clean deps install-tools fmt lint lint-new dev-setup test-all test-unit schemathesis install-hooks migrate-status migrate-validate river-migrate
+.PHONY: help tests tests-coverage build build-backfill-embeddings run run-backfill-embeddings init-db clean docker-up docker-down docker-clean deps install-tools fmt lint lint-new dev-setup test-all test-unit schemathesis install-hooks migrate-status migrate-validate river-migrate
```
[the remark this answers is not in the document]
xernobyl: Fixed.

Makefile
```diff
@@ -62,6 +63,10 @@ build-backfill-embeddings:
 	go build -o bin/backfill-embeddings ./cmd/backfill-embeddings
 	@echo "Binary created: bin/backfill-embeddings"
 
+# Run the backfill-embeddings command (loads .env for DATABASE_URL etc.)
+run-backfill-embeddings:
+	@(set -a && [ -f .env ] && . ./.env && set +a; go run ./cmd/backfill-embeddings)
+
```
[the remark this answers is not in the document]
xernobyl: Fixed.

migrations/005_embeddings_indexes_and_feedback_tenant_required.sql
```diff
@@ -0,0 +1,25 @@
...
+
+-- +goose Down
+DROP INDEX CONCURRENTLY IF EXISTS idx_embeddings;
+CREATE INDEX CONCURRENTLY idx_embeddings ON embeddings USING hnsw (embedding halfvec_cosine_ops);
+
+ALTER TABLE feedback_records ALTER COLUMN tenant_id DROP NOT NULL;
+UPDATE feedback_records SET tenant_id = NULL WHERE tenant_id = 'migrated-null-tenant';
+ALTER TABLE feedback_records DROP COLUMN IF EXISTS tenant_id_was_null;
+
+DROP INDEX IF EXISTS idx_embeddings_model;
```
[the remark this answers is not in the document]
xernobyl: Fixed.

internal/config/config.go
```diff
@@ -93,6 +96,24 @@ func getEnvAsInt(key string, defaultValue int) int {
 	return value
 }
 
+// GetEnvAsBool retrieves an environment variable as a boolean. "true", "1", "yes" (case-insensitive) => true; else false.
+// Exported so other cmd packages (e.g. backfill-embeddings) can reuse it without duplicating logic.
```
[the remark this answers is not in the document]
xernobyl: Fixed.

internal/service/embedding_provider_test.go
```diff
@@ -142,3 +142,70 @@ func TestEmbeddingProvider_PublishEvent_FeedbackRecordUpdated_valueTextInChanged
...
+	t.Run("e5 returns passage prefix", func(t *testing.T) {
+		assert.Equal(t, "passage: ", EmbeddingPrefixForProvider("e5"))
+	})
+	t.Run("unknown returns empty", func(t *testing.T) {
+		assert.Empty(t, EmbeddingPrefixForProvider("unknown"))
+	})
+	t.Run("case insensitive", func(t *testing.T) {
+		assert.Equal(t, "search_document: ", EmbeddingPrefixForProvider("NOMIC"))
+	})
+}
```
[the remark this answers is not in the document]
xernobyl: Fixed.

pkg/embeddings/normalize.go
```diff
@@ -0,0 +1,30 @@
...
+		return
+	}
+
+	// 2. Find the square root (the magnitude/length of the vector)
+	magnitude := math.Sqrt(sumSquares)
+
+	// 3. Divide each dimension by the magnitude
+	for i := range vector {
+		vector[i] = float32(float64(vector[i]) / magnitude)
+	}
```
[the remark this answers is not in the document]
xernobyl: Fixed.

internal/service/search_service.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		s.logger.Error("semantic search: nearest failed", "error", err, "model", s.model)
+
+		return out, fmt.Errorf("nearest feedback records: %w", err)
+	}
+
+	out.Results = results
+	if len(results) == limit {
+		last := results[len(results)-1]
+
```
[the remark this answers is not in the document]
xernobyl: Fixed.

internal/service/search_service.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/formbricks/hub/internal/repository"
+)
+
+const searchQueryEmbeddingCacheName = "search_query_embedding"
+
+// Sentinel errors for search (used by handlers for status mapping).
+var (
+	ErrMissingTenantID   = errors.New("environment_id is required")
+	ErrEmptyQuery        = errors.New("query is required and must be non-empty")
+	ErrEmbeddingNotFound = repository.ErrEmbeddingNotFound
```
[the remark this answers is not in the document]
xernobyl: Fixed.

internal/service/embedding_provider.go
```diff
@@ -131,19 +136,88 @@ func recordIDFromEventData(data any) uuid.UUID {
...
+	case "nomic":
+		return "search_document: "
+	case "e5", "intfloat/multilingual-e5-large", "intfloat/e5-large":
+		return "passage: "
+	case "mistral":
+		return "search_document: "
+	default:
+		return ""
+	}
+}
```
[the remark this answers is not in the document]
xernobyl: Fixed.

openapi.yaml
```diff
@@ -544,6 +551,173 @@ paths:
...
+                    default: 10
+                    minimum: 1
+                    maximum: 100
+                - name: offset
+                  in: query
+                  description: Number of results to skip (OFFSET-based paging). Ignored if cursor is set. Capped at 1000 for performance.
+                  schema:
+                    type: integer
+                    minimum: 0
+                    default: 0
```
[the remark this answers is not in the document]
xernobyl: Fixed.

migrations/005_embeddings_indexes_and_feedback_tenant_required.sql
```diff
@@ -0,0 +1,25 @@
+-- +goose NO TRANSACTION
+-- +goose Up
+-- B-tree index on model to help filter embeddings by model before/while using the HNSW vector index.
+CREATE INDEX idx_embeddings_model ON embeddings (model);
```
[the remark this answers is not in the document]
xernobyl: Fixed.

internal/service/search_service.go
```diff
@@ -0,0 +1,245 @@
...
+	if len(results) == limit {
+		last := results[len(results)-1]
+
+		nextCursor, err := EncodeSearchCursor(1-last.Score, last.FeedbackRecordID)
+		if err != nil {
+			return out, fmt.Errorf("encode next cursor: %w", err)
+		}
+
+		out.NextCursor = nextCursor
+	}
```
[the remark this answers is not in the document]
xernobyl: Fixed.

PR #46, opened 2026-03-06
chore: production hardening — Go 1.26, security (SSRF, hide signing key), config, pagination

xernobyl:
## What does this PR do?

Production hardening: CI upgrades, security hardening, config validation, and hiding webhook signing keys.

### 1. CI and tooling
- Upgrade to Go 1.26 and golangci-lint for compatibility
- Update tool versions (goose, River, etc.) in Makefile, Dockerfile, workflows
- Add test timeout for `make tests`
- Scope gosec suppressions (G704/G706/G101) to test files

### 2. Config and security
- Add/use database pool options (max/min conns, lifetime, idle, health check, connect timeout)
- SSRF hardening: reject private/loopback/link-local IPs for webhook URLs; optional host blacklist
- Config validation (e.g. embedding provider/model)

### 3. Pagination
- Enforce pagination invariant: return `ErrPaginationInvariantViolated` when `hasMore` is true but `encodeLast` is nil
- Align test expectations with repository ordering

### 4. Hide webhook signing key (security)
- Omit `signing_key` from GET, List, and Update responses
- Add `WebhookPublic` (no `signing_key`), `ToWebhookPublic()`, and `ListWebhooksPublicResponse`
- Update OpenAPI with `WebhookPublicData` and point GET/PATCH webhook and List webhooks responses to it
- Create response still returns full webhook (including `signing_key`) so clients can store it once

### 5. Other
- Replace `ptrFloat64` with Go 1.22+ `new()` for pointer fields
- Improve tests (deterministic ordering, SSRF validation coverage)
- Fix invalid `new()` usage in tests

### Review Provenance
- This PR includes AI-assisted review content.
- The posted review comments were human-reviewed and approved before submission.

**Base:** `main` (or `chore/ci_config_security_new_pattern` if that is the PR base)

## How should this be tested?

- `make build`
- `make tests` — all integration tests pass
- `make fmt` and `make lint` — no new warnings
- Verify webhook endpoints:
  - `GET /v1/webhooks/{id}` — response omits `signing_key`
  - `GET /v1/webhooks` — list items omit `signing_key`
  - `PATCH /v1/webhooks/{id}` — response omits `signing_key`
  - `POST /v1/webhooks` — response includes `signing_key`
- (Optional) Test SSRF rejection with loopback/private IP webhook URLs

### Required
- [x] Filled out the "How to test" section
- [x] Read Repository Guidelines
- [x] Self-reviewed
- [x] Ran `make build`
- [x] Ran `make tests`
- [ ] Merged latest from base branch
- [x] No schema changes (no migrations)

### Appreciated
- [x] Updated OpenAPI spec for webhook responses
- [x] API behavior changes documented above
- [ ] Updated docs in `docs/` if applicable
- [ ] Ran `make tests-coverage` for new logic


internal/config/config.go
```diff
@@ -176,6 +214,74 @@ func Load() (*Config, error) {
...
+	}
+
+	databaseHealthCheckPeriod := time.Duration(databaseHealthCheckPeriodSecs) * time.Second
+
+	databaseConnectTimeoutSecs := getEnvAsInt("DATABASE_CONNECT_TIMEOUT_SECONDS", defaultDatabaseConnectTimeoutSeconds)
+	if databaseConnectTimeoutSecs <= 0 {
+		databaseConnectTimeoutSecs = defaultDatabaseConnectTimeoutSeconds
+	}
+
+	databaseConnectTimeout := time.Duration(databaseConnectTimeoutSecs) * time.Second
```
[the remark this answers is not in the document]
xernobyl: fixed.

internal/service/webhooks_service_test.go
```diff
@@ -64,7 +64,7 @@ func (noopPublisher) PublishEventWithChangedFields(_ context.Context, _ datatype
 
 func TestWebhooksService_CreateWebhook_InvalidSigningKey(t *testing.T) {
 	ctx := context.Background()
-	svc := NewWebhooksService(&mockWebhooksRepo{count: 0}, noopPublisher{}, 10)
+	svc := NewWebhooksService(&mockWebhooksRepo{count: 0}, noopPublisher{}, 10, nil)
```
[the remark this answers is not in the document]
xernobyl: Fixed.

.golangci.yml
```diff
@@ -43,6 +43,10 @@ linters:
         linters:
           - nilnil       # mocks often return (nil, nil) for success
           - err113       # mocks and test helpers use dynamic errors
+      # gosec false positives: G706 log injection (slog key-values), G704 SSRF (tests + validated webhook URLs), G101 test DB credential
+      - linters:
+          - gosec
+        text: "G70[46]|G101"
```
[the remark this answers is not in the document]
xernobyl: Fixed.

internal/service/webhooks_service.go
```diff
@@ -99,6 +110,45 @@ func validateSigningKey(key string) error {
...
+			addr.IsLinkLocalMulticast() || addr.IsUnspecified() {
+			return huberrors.NewValidationError("url", "webhook URL host is not allowed (private/internal)")
+		}
+	}
+
+	if _, blocked := blacklist[host]; blocked {
+		return huberrors.NewValidationError("url", "webhook URL host is not allowed (blacklisted)")
+	}
+
+	return nil
```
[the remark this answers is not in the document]
xernobyl: Fixed.

Makefile
```diff
@@ -200,10 +200,10 @@ deps:
 
 # Install development tools
 # Tool versions - update these periodically
-GOLANGCI_LINT_VERSION := v2.8.0
+GOLANGCI_LINT_VERSION := v2.10.1
 GOVULNCHECK_VERSION := v1.1.4
-GOOSE_VERSION := v3.26.0
-RIVER_VERSION := v0.30.2
+GOOSE_VERSION := v3.27.0
+RIVER_VERSION := v0.31.0
```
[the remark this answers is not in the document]
xernobyl: Fixed.

internal/config/config.go
```diff
@@ -199,6 +305,20 @@ func Load() (*Config, error) {
 		MessagePublisherPerEventTimeout: time.Duration(perEventTimeoutSecs) * time.Second,
 		ShutdownTimeout:                 time.Duration(shutdownTimeoutSecs) * time.Second,
 		WebhookMaxCount:                 webhookMaxCount,
+		WebhookHTTPTimeout:              webhookHTTPTimeout,
+
+		WebhookEnqueueMaxRetries:     webhookEnqueueMaxRetries,
```
mattinannt, line 310: `WEBHOOK_ENQUEUE_MAX_RETRIES`, `WEBHOOK_ENQUEUE_INITIAL_BACKOFF_MS`, and `WEBHOOK_ENQUEUE_MAX_BACKOFF_MS` are loaded here, but no runtime path consumes them yet. `WebhookProvider.PublishEvent` still calls `InsertMany` once per batch and returns immediately on the first error, so these become dead configuration despite being documented in `.env.example`.

Recommended fix: thread these values into `NewWebhookProvider` and wrap `InsertMany` in a bounded retry loop with exponential backoff plus jitter, for example: `for attempt := 0; attempt <= maxRetries; attempt++ { _, err = inserter.InsertMany(...); if err == nil { break }; if attempt == maxRetries { return }; sleep(backoffWithJitter(attempt)) }`. If that behavior is out of scope for this PR, remove these env vars from config and docs for now so operators are not given knobs that do nothing.

AI-assisted review, human-reviewed and approved before posting.

internal/service/webhook_sender.go
```diff
@@ -33,26 +34,56 @@ type WebhookSender interface {
...
+					return nil, fmt.Errorf("invalid address %q: %w", addr, err)
+				}
+
+				allowed, err := resolveWebhookHost(ctx, host, urlHostBlacklist)
+				if err != nil {
+					return nil, err
+				}
+
+				// Dial the first validated IP to prevent DNS rebinding.
+				target := net.JoinHostPort(allowed[0].String(), port)
```
mattinannt: Dialing only `allowed[0]` changes behavior from Go's default address fallback to a single-address pin. That can produce false delivery failures for webhook hosts behind CDNs, dual-stack DNS, or partially unhealthy pools, and repeated failures here can disable an otherwise healthy webhook after max attempts.

Recommended fix: keep the SSRF validation step, but preserve multi-address dialing instead of pinning the first IP. A practical approach is to resolve and validate the full address set, then try each validated IP in order until one connects, returning the last error only if they all fail. Another good option is to validate in a custom resolver path and still let the transport attempt multiple validated addresses.

AI-assisted review, human-reviewed and approved before posting.

internal/service/webhook_sender.go
```diff
@@ -33,26 +34,56 @@ type WebhookSender interface {
...
+
+				return dialer.DialContext(ctx, network, target)
+			},
+		}
+
+		httpClient = &http.Client{
+			Transport: transport,
+			Timeout:   webhookHTTPTimeout,
```
mattinannt: This still uses the hard-coded `webhookHTTPTimeout`, so the new `WEBHOOK_HTTP_TIMEOUT_SECONDS` setting remains dead even though the PR parses and documents it. The worker timeout is also still fixed separately, which means operators cannot actually tune delivery timeout behavior from config right now.

Recommended fix: pass `cfg.WebhookHTTPTimeout` into `NewWebhookSenderImpl` and use it here instead of the constant, then derive `WebhookDispatchWorker.Timeout()` from the same config (for example `timeout + 5s`) so the HTTP client timeout and River job timeout stay aligned. If configurable timeouts are not intended in this PR, remove the env var and docs to avoid a misleading contract.

AI-assisted review, human-reviewed and approved before posting.

mattinannt (changes requested):
AI-assisted review, human-reviewed and approved before posting. Requesting changes for three issues that introduce dead configuration or make webhook delivery less reliable in production:

1. `WEBHOOK_ENQUEUE_*` is parsed and documented but not wired into any retry path.
2. `WEBHOOK_HTTP_TIMEOUT_SECONDS` is parsed and documented but the sender still uses a hard-coded timeout.
3. The new SSRF-safe dialer now pins delivery to the first resolved IP, which can fail healthy multi-address webhook endpoints.

Recommended direction: either fully wire the new config through the runtime behavior in this PR, or remove the new knobs until the behavior exists. For the dialer, keep the SSRF validation but preserve multi-address connection attempts instead of hard-pinning `allowed[0]`.

internal/service/webhook_sender.go
xernobyl: Fixed.

internal/service/webhook_sender.go
xernobyl: Fixed.

internal/config/config.go
xernobyl: Fixed.

mattinannt (approved):

PR #49, opened 2026-03-10
fix: add uuidv7() function definition to migration 001

BhagyaAmarasinghe:
## Summary
- The `uuidv7()` function is used as a column default in migrations 001 (`feedback_records.id`) and 002 (`webhooks.id`) but was never defined in any migration — it was manually created on each database
- This breaks fresh deployments and schema resets (e.g. the 0.2.0 staging upgrade) because goose cannot create the tables
- Adds the standard PL/pgSQL UUIDv7 implementation right after the `pgcrypto` extension it depends on, and drops it in the down migration

## Test plan
- [ ] CI migration validation passes
- [ ] Fresh `goose up` against an empty database succeeds (all 6 migrations)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/migrations/001_initial_schema.sql
+++ b/migrations/001_initial_schema.sql
@@ -5,6 +5,21 @@
 CREATE EXTENSION IF NOT EXISTS vector;
 CREATE EXTENSION IF NOT EXISTS "pgcrypto";
 
+-- UUIDv7: time-ordered UUIDs (PG 16 and below; native in PG 17+).
+-- Requires pgcrypto for gen_random_bytes().
+CREATE OR REPLACE FUNCTION uuidv7() RETURNS uuid AS $fn$
+DECLARE
+  unix_ts_ms bytea;
+  uuid_bytes bytea;
+BEGIN
+  unix_ts_ms = substring(int8send(floor(extract(epoch from clock_timestamp()) * 1000)::bigint) from 3);
+  uuid_bytes = unix_ts_ms || gen_random_bytes(10);
+  uuid_bytes = set_byte(uuid_bytes, 6, (b'0111' || get_byte(uuid_bytes, 6)::bit(4))::bit(8)::int);
+  uuid_bytes = set_byte(uuid_bytes, 8, (b'10' || get_byte(uuid_bytes, 8)::bit(6))::bit(8)::int);
+  RETURN encode(uuid_bytes, 'hex')::uuid;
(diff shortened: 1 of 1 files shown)


migrations/001_initial_schema.sql
```diff
@@ -5,6 +5,23 @@
...
+  uuid_bytes bytea;
+BEGIN
+  unix_ts_ms = substring(int8send(floor(extract(epoch from clock_timestamp()) * 1000)::bigint) from 3);
+  uuid_bytes = unix_ts_ms || gen_random_bytes(10);
+  uuid_bytes = set_byte(uuid_bytes, 6, (b'0111' || get_byte(uuid_bytes, 6)::bit(4))::bit(8)::int);
+  uuid_bytes = set_byte(uuid_bytes, 8, (b'10' || get_byte(uuid_bytes, 8)::bit(6))::bit(8)::int);
+  RETURN encode(uuid_bytes, 'hex')::uuid;
+END
+$$ LANGUAGE plpgsql VOLATILE;
+-- +goose StatementEnd
```
xernobyl, line 23: Make sure the function doesn't replace the existing uuidv7 function for newer (>=PG18) versions, with something like (not tested):

```
DO $mig$
BEGIN
  IF NOT EXISTS (
    SELECT 1 FROM pg_proc p
    JOIN pg_namespace n ON p.pronamespace = n.oid
    WHERE n.nspname = 'public' AND p.proname = 'uuidv7'
  ) THEN
    CREATE FUNCTION uuidv7() RETURNS uuid AS $fn$
    DECLARE
      unix_ts_ms bytea;
      uuid_bytes bytea;
    BEGIN
      unix_ts_ms = substring(int8send(floor(extract(epoch from clock_timestamp()) * 1000)::bigint) from 3);
      uuid_bytes = unix_ts_ms || gen_random_bytes(10);
      uuid_bytes = set_byte(uuid_bytes, 6, (b'0111' || get_byte(uuid_bytes, 6)::bit(4))::bit(8)::int);
      uuid_bytes = set_byte(uuid_bytes, 8, (b'10' || get_byte(uuid_bytes, 8)::bit(6))::bit(8)::int);
      RETURN encode(uuid_bytes, 'hex')::uuid;
    END
    $fn$ LANGUAGE plpgsql VOLATILE;
  END IF;
END
$mig$;
```

PR #64, opened 2026-04-27
feat: support UUIDv7 on Postgres 16 and 17

BhagyaAmarasinghe:
## Summary
This change makes Hub bootstrap cleanly on PostgreSQL 16 and 17 while preserving native UUIDv7 behavior on PostgreSQL 18.

## What changed
- add a SQL `public.uuidv7()` fallback in `001_initial_schema.sql` so fresh installs on PG16/17 can create tables that use `DEFAULT uuidv7()`
- keep the default expression unchanged so PG18 continues to resolve to `pg_catalog.uuidv7()`
- switch the local compose default to `pg17`
- add PG16/PG17 compatibility jobs to CI while keeping the existing PG18 integration and coverage flow
- assert UUIDv7 shape in integration tests and replace `example.com` webhook URLs with public IP literal URLs so compatibility checks do not depend on external DNS

## Why this approach
Hub currently relies on database-side UUID generation for persisted record and webhook IDs, so the failure mode is migration-time schema creation rather than Go runtime behavior.

I kept the fix in the bootstrap migration instead of moving ID generation into application code because:
- it preserves the existing insert paths and public API behavior
- it avoids widening the change surface into repositories and services
- it lets older Postgres versions bootstrap without extra operator steps or external extensions
- it keeps PG18 on the native implementation automatically

I also added explicit compatibility jobs rather than only changing local compose defaults, because without CI coverage this would regress easily and the original failure would remain invisible until integration into older XM Suite installations.

## Root cause
`feedback_records.id` and `webhooks.id` use `DEFAULT uuidv7()`, but PostgreSQL 16 and 17 do not provide that built-in function. The schema only worked because local and CI environments were pinned to PG18.

## Validation
- `PATH=/Users/bhagya/work/bin:$PATH make migrate-validate`
- `go test ./internal/...`
- fresh bootstrap + River migrations + targeted integration tests on `pgvector/pgvector:pg16`
- fresh bootstrap + River migrations + full `./tests/...` suite on `pgvector/pgvector:pg17`
- fresh bootstrap + River migrations + targeted integration tests on `pgvector/pgvector:pg18`

Closes ENG-755.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,8 +49,76 @@ jobs:
       - name: Run unit tests
         run: make test-unit
 
+  compatibility-tests:
+    name: Compatibility Tests (${{ matrix.postgres_image }})
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+
+    strategy:
+      fail-fast: false
+      matrix:
+        postgres_image: [pg16, pg17]
+
+    services:
+      postgres:
+        image: pgvector/pgvector:${{ matrix.postgres_image }}
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: test_db
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    env:
+      DATABASE_URL: postgres://postgres:postgres@localhost:5432/test_db?sslmode=disable
+      API_KEY: test-api-key-for-ci
+
+    steps:
+      - name: Checkout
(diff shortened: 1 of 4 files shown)


xernobyl (changes requested):
Looks good, just fix the coderabbit review where it complains about you skipping the built-in uuidv7 function in >= pg18.

xernobyl:
Out of curiosity I decided to do a benchmark. Wondering if there was some overhead related to the new function, but the results seem to still justify it :)

image

BhagyaAmarasinghe:
> Out of curiosity I decided to do a benchmark. Wondering if there was some overhead related to the new function, but the results seem to still justify it :)
> 
> image

 It confirms the split we landed on: use native uuidv7() on PG18 and keep the SQL fallback only for older Postgres versions.

xernobyl (approved):

PR #65, opened 2026-04-28
fix: enforce tenant scope on webhooks

xernobyl:
## Summary

Fixes ENG-796 by enforcing tenant isolation for webhook dispatch end to end.

- Require `tenant_id` for webhook creation and prevent clearing it on update.
- Disable legacy null/blank-tenant webhooks and add a database constraint preventing future null/blank webhook tenants.
- Replace broad webhook fan-out with tenant-exact lookup by event type and tenant.
- Carry tenant scope in River webhook dispatch args and re-check it in the worker before side effects.
- Reject malformed webhook jobs where job-level `tenant_id` conflicts with payload `tenant_id`.
- Keep public webhook deleted-event `data` as an ID array while using tenant-aware internal delete event data.
- Require `tenant_id` for feedback bulk delete so delete events cannot fan out across tenants.
- Update `AGENTS.md` with access-path parity rules to prevent future tenant leaks through alternate paths.

## Root Cause

Webhook dispatch treated `tenant_id` like an optional filter/global fallback. Feedback records became tenant-required, but the webhook fan-out path could still select enabled webhooks without enforcing the same tenant boundary, so async dispatch was not equivalent to the primary tenant-scoped data paths.

## User Impact

Tenant-scoped webhooks now only receive events for their exact tenant. Missing or conflicting tenant context skips dispatch and logs the problem instead of sending cross-tenant data.

## Migration

Adds `008_require_webhook_tenant_id.sql`:

- disables existing webhooks with null or blank `tenant_id`
- records a disabled reason/timestamp
- adds a check constraint preventing future null or blank webhook tenants

No new index is required for this change; existing webhook tenant/event/enabled indexes already cover the dispatch query shape.

## Testing

- `go test ./...` ✅
- `golangci-lint run ./...` ✅
- `make migrate-validate` ✅
- `make lint-openapi` ✅
- Changed production line coverage: `133/164` lines, `81.1%` ✅

Pre-commit also ran migration validation, fmt, lint, and unit tests successfully, then failed on a missing local Make target `check-coverage-staged`; the commit was created with `--no-verify` after the equivalent checks and coverage measurement passed manually.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -29,10 +29,13 @@
 
 - Treat `tenant_id` as a security boundary, not a convenience filter. Tenant-owned data must never be read, enqueued, dispatched, cached, searched, embedded, or deleted across tenants.
 - When making a model, migration, API request, or repository change involving tenant-owned data, audit every downstream path that carries or derives from that data: handlers, services, repositories, message publishers, River job args, workers, webhook payloads, search, embeddings, bulk operations, logs, and metrics.
-- Prefer tenant-aware repository/service methods for tenant-owned workflows. Avoid adding broad helpers that return all enabled/all matching resources when the caller is dispatching, processing, or exposing tenant data.
+- Tenant access rules must be consistent across every path that can observe, mutate, derive from, or act on the same resource. If one API endpoint, repository method, search path, webhook dispatch path, worker, backfill, bulk operation, or export path requires tenant scope, every alternate path for that resource must enforce the same tenant boundary.
+- Do not model `tenant_id` as an optional filter for tenant-owned resources. Prefer required tenant parameters in service/repository method signatures (`tenantID string`, not `*string`) unless the domain explicitly supports global resources and documents that behavior.
+- Prefer tenant-aware repository/service methods for tenant-owned workflows. Avoid adding broad helpers that return all enabled/all matching resources when the caller is dispatching, processing, deriving, exporting, or exposing tenant data.
 - Async jobs must carry the tenant boundary when the source data has one, and workers must re-check tenant scope before doing side effects. Do not rely only on enqueue-time filtering.
-- Global resources may intentionally have `tenant_id = NULL`, but code must make that behavior explicit. A missing tenant on event data should not match tenant-scoped resources.
-- For any tenant-scoping change, include verification at the boundary where the leak could happen: database query behavior, service fan-out, worker execution, and API behavior when relevant. Tests are the evidence; the invariant belongs in the architecture.
+- Global resources may intentionally have `tenant_id = NULL` only when the domain explicitly documents them as non-tenant-owned. Webhooks are tenant-owned and must require a non-empty `tenant_id`; a missing tenant on event data must not match any webhook.
+- When changing access rules for a tenant-owned model, search for all alternate access paths by resource name and by derived side effects: list, get, update, delete, bulk delete, webhook fan-out, River jobs, workers, embeddings, search, exports, cache invalidation, logs, and metrics.
+- For any tenant-scoping change, include verification at the boundary where the leak could happen: database query behavior, service fan-out, worker execution, and API behavior when relevant. Include at least one alternate-path regression test proving that data allowed through the primary path cannot leak through async dispatch, bulk operations, derived indexes, exports, or background workers. Tests are the evidence; the invariant belongs in the architecture.
 
 ## Testing Guidelines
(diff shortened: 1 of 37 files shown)


AGENTS.md
```diff
@@ -25,6 +25,14 @@
 - Prefer Go naming conventions (CamelCase for exported, lowerCamel for unexported).
 - Keep package names short and domain-focused (e.g., `repository`, `service`).
 
+## Multi-Tenancy & Data Isolation
+- Treat `tenant_id` as a security boundary, not a convenience filter. Tenant-owned data must never be read, enqueued, dispatched, cached, searched, embedded, or deleted across tenants.
```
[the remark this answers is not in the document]
xernobyl: Fixed.

internal/service/webhooks_service.go
```diff
@@ -369,14 +376,58 @@ func (s *WebhooksService) UpdateWebhook(ctx context.Context, id uuid.UUID, req *
...
+
+	*tenantID = normalized
+
+	return nil
+}
+
 // DeleteWebhook deletes a webhook by ID.
-// Publishes WebhookDeleted with data = [id] (array of deleted IDs) for consistency with feedback record deletes.
+// Publishes WebhookDeleted with tenant-aware deleted IDs.
 func (s *WebhooksService) DeleteWebhook(ctx context.Context, id uuid.UUID) error {
```
BhagyaAmarasinghe, line 411: DeleteWebhook publishes webhook.deleted using a separate pre-delete read. Because tenant_id is mutable, concurrent update+delete can emit the delete event to the wrong tenant. DELETE ... RETURNING tenant_id would avoid that race.

internal/workers/webhook_dispatch_test.go
```diff
@@ -75,7 +82,9 @@ func TestWebhookDispatchWorker_Work(t *testing.T) {
 	})
 
 	t.Run("returns nil on send success", func(t *testing.T) {
-		repo := &mockDispatchRepo{webhook: &models.Webhook{ID: webhookID, Enabled: true, URL: "http://x", SigningKey: "sk"}}
+		repo := &mockDispatchRepo{
+			webhook: &models.Webhook{ID: webhookID, Enabled: true, URL: "http://x", SigningKey: "sk", TenantID: &tenantID},
+		}
```
BhagyaAmarasinghe, line 87: The new tenant gate breaks compatibility with already-queued legacy delete jobs. Old River jobs have neither args.TenantID nor data.tenant_id, so pending feedback_record.deleted / webhook.deleted deliveries will be silently skipped after rollout.

internal/service/feedback_records_service.go
```diff
@@ -182,36 +183,72 @@ func (s *FeedbackRecordsService) UpdateFeedbackRecord(
...
+// BulkDeleteFeedbackRecords deletes all feedback records matching user_id.
+// When tenant_id is provided, deletion is restricted to that tenant; otherwise all user records are deleted.
+// It publishes one tenant-aware FeedbackRecordDeleted event per tenant represented in the deleted rows.
+func (s *FeedbackRecordsService) BulkDeleteFeedbackRecords(ctx context.Context, filters *models.BulkDeleteFilters) (int, error) {
+	if filters == nil || filters.UserID == "" {
 		return 0, ErrUserIDRequired
 	}
 
-	ids, err := s.repo.BulkDelete(ctx, userID, tenantID)
+	if filters.TenantID != nil {
```
BhagyaAmarasinghe, line 223: Bulk delete now trims tenant_id, but feedback-record creation still stores raw tenant strings. That makes padded tenants undeletable via the scoped endpoint and prevents them from matching tenant-scoped webhooks.

BhagyaAmarasinghe (changes requested):

internal/service/feedback_records_service.go
xernobyl: Addressed.

internal/workers/webhook_dispatch_test.go
xernobyl: I don't think it's an issue to break compatibility with already-queued legacy delete jobs, since no one is using webhooks in production.

internal/service/webhooks_service.go
xernobyl: Addressed.

BhagyaAmarasinghe (approved):

PR #72, opened 2026-05-07
fix: allow staging hub image tag

BhagyaAmarasinghe:
What changed:
- Adds a manual ECR workflow option to tag a Hub build as staging.
- Allows staging to be rebuilt from current main with both /app/hub-api and /app/hub-worker in the image.

Tested:
- git diff --check
- Dispatched Build & Push to AWS ECR with version_override=0.2.1-staging.20260507 and deploy_staging=true.
- Restarted hub-stage-worker and verified the rollout succeeded.

Notes:
- The previous live hub:staging image did not include /app/hub-worker, which caused hub-stage-worker to crash-loop.

### DIFF
--- a/.github/workflows/build-and-push-ecr.yml
+++ b/.github/workflows/build-and-push-ecr.yml
@@ -16,6 +16,11 @@ on:
         required: false
         default: false
         type: boolean
+      deploy_staging:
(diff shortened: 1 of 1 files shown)


xernobyl (approved):
LGTM

PR #86, opened 2026-06-12
fix: preserve deep taxonomy tree children

BhagyaAmarasinghe:
## Summary
- Fix taxonomy tree reconstruction so deep descendants are attached recursively instead of losing grandchildren when nodes are copied by value.
- Add a regression test that builds a 5-level taxonomy path and verifies sorted level-4 leaves are preserved.

## Why
The Formbricks taxonomy explorer depends on Hub returning the full tree. The previous reconstruction appended value copies before descendants were attached, so only shallow levels were visible in the UI.

## Validation
- `go test ./internal/repository`


xernobyl (approved):
Fix the linting errors, other than that LGTM.

PR #89, opened 2026-06-22
feat: add tenant settings for language enrichment

xernobyl:
## What does this PR do?

First task in the **Language consolidation** chain (project: Hub Native Enrichment). Adds a tenant-scoped settings store so language enrichment can be configured per tenant — starting with `target_language` (a normalized BCP-47 locale, e.g. `en-US`) — and, following review, wires the **consumption side** so the setting is usable end to end (see _Added during review_ below). Pairs with ENG-1255 (translate feedback records), which reads these settings.

Fixes [ENG-1254](https://linear.app/formbricks/issue/ENG-1254)

**Design**
- **Storage:** one row per tenant in a new `tenant_settings` table, keyed by `tenant_id` (natural PK, no surrogate id), with a single open-ended `settings` JSONB column mapped to a typed `EnrichmentSettings` struct — new settings are added as struct fields with **no schema migration**.
- **API:** `GET`/`PUT`/`PATCH /v1/tenants/{tenant_id}/settings`. `PUT` is a full replace. `PATCH` is an **RFC 7396 JSON Merge Patch** (`application/merge-patch+json`; `application/json` also accepted): a member with a value sets that setting, an explicit JSON `null` removes the key, and an omitted member is left unchanged — and it creates the row when the tenant has none. An empty string is rejected (a value must be a valid locale; clear via `null`). Both verbs normalize the locale via `golang.org/x/text/language` (`en-us` → `en-US`). An unconfigured tenant returns **200 with empty settings, not 404** — consumers decide the fallback.
- **Body cap:** `PUT` and `PATCH` bodies are capped at 8 KiB (`http.MaxBytesReader`); an oversized body returns **413** with a dedicated `content_too_large` problem code (added to the shared RFC 9457 response layer + the OpenAPI `ErrorModel` enum).
- **Tenant scoping:** `tenant_id` is path-only (the body has no `tenant_id`; unknown fields are rejected); reads are `WHERE tenant_id = $1`; writes are `INSERT … ON CONFLICT (tenant_id)` behind the existing shared tenant write-lock. The tenant-data purge now also deletes `tenant_settings`.
- **Reuse:** the tenant write-lock helpers, tenant-id normalization, the `x/text` dependency, and the response/validation helpers — no new dependencies. `PUT` and `PATCH` share one body-decode/validate helper (`decodeSettingsBody`) and one locked-write helper (`writeSettings`). The merge patch is applied in SQL — `(settings || set) - removeKeys::text[]` — so `null` members delete keys while unmentioned keys survive; a small generic `Optional[T]` distinguishes absent / null / value when decoding the patch body.

**Trust boundary:** like every existing Hub endpoint, the service authenticates with a single static API key and the caller passes `tenant_id`; there is no per-tenant authn here — isolation is data-scoping (consistent with the ENG-1289 conclusion).

**Deferred to follow-ups (out of scope):**
- A repo-wide request **body-limit middleware** (the 8 KiB cap here is per-handler; other endpoints remain uncapped).
- Making the settings cache **multi-replica-coherent** (TTL/Redis): the eviction added below is per-process, so cross-replica staleness is still TTL-bounded.

**API examples**
```
PUT /v1/tenants/org-123/settings    {"target_language":"de-de"}
→ 200 {"tenant_id":"org-123","settings":{"target_language":"de-DE"}}

PATCH /v1/tenants/org-123/settings  {"target_language":"fr-fr"}   (merge; unspecified keys preserved)
→ 200 {"tenant_id":"org-123","settings":{"target_language":"fr-FR"}}

PATCH /v1/tenants/org-123/settings  {"target_language":null}      (RFC 7396 null removes the key)
→ 200 {"tenant_id":"org-123","settings":{}}

GET /v1/tenants/never-configured/settings
→ 200 {"tenant_id":"never-configured","settings":{}}

PUT /v1/tenants/org-123/settings    (body > 8 KiB)
→ 413 {"code":"content_too_large","title":"Request Entity Too Large", ...}
```

## Added during review

Reviewer feedback (thanks @BhagyaAmarasinghe) surfaced gaps between a bare settings store and a working enrichment path. Addressed in this PR:

- **Deployment default language.** New `TRANSLATION_DEFAULT_LANGUAGE` env (BCP-47, canonicalized at load) — the fallback target for tenants with no `target_language` of their own; empty keeps translation strictly per-tenant opt-in. It is threaded through the service so a tenant's *effective* target — `COALESCE(its own target_language, TRANSLATION_DEFAULT_LANGUAGE)` — is resolved consistently at the enqueue gate, **both** backfills (global and per-tenant), and the `SetTranslation` write-guard. So an operator can turn translation on for every tenant with one env var; the guard persists a default-resolved write while still superseding a stale former-explicit-target write after a tenant clears its target; and clearing a tenant's target re-translates its existing records to the default.
- **Cache coherence.** A settings write now evicts the tenant's cached settings (composed with the translation backfill listener), so a changed or newly enabled target is seen by the enqueue gate immediately instead of after the cache TTL.
- **Stale translation clear.** An update to `value_text`/`language` now clears `value_text_translated`/`translation_lang_key` — but only on an actual change — so an edited record can't keep a stale translation; the row falls back to the original text and stays recoverable by a backfill.

**On ENG-1254's original contract:** the ticket sketched `translation_enabled` + `target_language` + env-default + explicit disabled-vs-default. We simplified to **env-based enablement** (`TRANSLATION_PROVIDER` + `TRANSLATION_MODEL` switch the whole deployment) **+ `target_language` + `TRANSLATION_DEFAULT_LANGUAGE` fallback** — no per-tenant `translation_enabled`. Opt-in is "the tenant has a target" (or the deployment default), which removes a redundant flag and the ambiguous enabled-but-no-target state. [ENG-1254](https://linear.app/formbricks/issue/ENG-1254) has been updated to this design.

## How should this be tested?

Config: `API_KEY=<key>`, `DATABASE_URL=postgres://postgres:postgres@localhost:5432/test_db?sslmode=disable`. To exercise translation end to end, also set `TRANSLATION_PROVIDER` + `TRANSLATION_MODEL` (provider creds) and optionally `TRANSLATION_DEFAULT_LANGUAGE`.

Automated:
- `make init-db && make river-migrate` (applies schema incl. `012_tenant_settings.sql`)
- `make tests` — integration tests in `tests/`: round-trip, defaults (200 not 404), locale normalization, **tenant A↔B isolation**, body-`tenant_id` rejected, invalid locale → 400 (PUT **and** PATCH), purge-removes-settings, **413 on an oversized body** (PUT and PATCH, asserts the `content_too_large` code), and the **RFC 7396 `PATCH`** paths — value-merge preserves other keys, `null` removes a key (siblings kept, key deleted from the JSONB), an empty string is rejected (400), and PATCH creates the row for a brand-new tenant. Plus an `Optional[T]` tri-state unit test (absent/null/value).
- **Default-language fallback** (added during review): config validation/canonicalization of `TRANSLATION_DEFAULT_LANGUAGE` (unit), the enqueue-gate fallback + per-tenant override (provider unit), the default-aware **global and per-tenant** backfills, and the write-guard — a default-resolved write persists, while a stale former-explicit-target write (after a tenant clears its target) is superseded (integration in `tests/`).
- `make tests-coverage` — new code: service 100%, handler 100%, `Optional` 100%, repository file ~85%

Manual smoke (server on `:8099`):
```
curl -X PUT -H "Authorization: Bearer $API_KEY" -H 'Content-Type: application/json' \
  -d '{"target_language":"de-de"}' localhost:8099/v1/tenants/org-123/settings   # 200, de-DE
curl -X PATCH -H "Authorization: Bearer $API_KEY" -H 'Content-Type: application/merge-patch+json' \
  -d '{"target_language":"fr-fr"}' localhost:8099/v1/tenants/org-123/settings   # 200, fr-FR (merge)
curl -X PATCH -H "Authorization: Bearer $API_KEY" -H 'Content-Type: application/merge-patch+json' \
  -d '{"target_language":null}'   localhost:8099/v1/tenants/org-123/settings    # 200, {} (removed)
curl -H "Authorization: Bearer $API_KEY" localhost:8099/v1/tenants/unset/settings # 200, {}
curl localhost:8099/v1/tenants/org-123/settings                                   # 401 (no key)
```

### Required

- [x] Commented on my code in hard-to-understand bits (tenant write-lock rationale, defaults-when-unset, the `//nolint` for `ErrNoRows`, the RFC 7396 set/remove SQL, the default-language write-guard)
- [x] Ran `make build`
- [x] Ran `make tests` (integration tests in `tests/`, against `test_db`)
- [x] Ran `make fmt` and `make lint`; no new warnings (0 issues)
- [ ] Merged the latest changes from main onto my branch with `git pull origin main` — branched off `origin/main`@4141d58; rebase before merge
- [x] If database schema changed: added migration in `migrations/` with goose annotations and ran `make migrate-validate` (012_tenant_settings.sql; also applied via `goose up`)

### Appreciated

- [x] If API changed: added or updated OpenAPI spec and ran contract tests (`make tests`)
- [x] Updated docs in `docs/` if changes were necessary — companion hub-api-docs PRs (stainless-sdks/hub-api-docs#8 tenant settings, stainless-sdks/hub-api-docs#9 translated feedback incl. the `TRANSLATION_DEFAULT_LANGUAGE` env); the API reference is auto-generated from the OpenAPI spec


migrations/012_tenant_settings.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
```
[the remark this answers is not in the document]
xernobyl: Fixed in e76442a — switched to lowercase `-- +goose up` / `-- +goose down`. Confirmed this is the documented convention in `AGENTS.md` (lines 8 and 19: "use `-- +goose up` / `-- +goose down` annotations") and it matches the most recent migrations (009–011). Style-only since goose treats the keywords case-insensitively: `make migrate-validate` passes and the applied schema is unchanged.

tests/tenant_settings_test.go
```diff
@@ -0,0 +1,349 @@
...
+
+func TestTenantSettings_PatchBodyTooLargeRejected(t *testing.T) {
+	server, cleanup := setupTestServer(t)
+	defer cleanup()
+
+	oversized := `{"target_language":"` + strings.Repeat("a", 9000) + `"}`
+	resp := settingsRequest(t, server.URL, http.MethodPatch, testTenantID("patch-toobig"), oversized, true)
+	require.NoError(t, resp.Body.Close())
+	assert.Equal(t, http.StatusRequestEntityTooLarge, resp.StatusCode)
+}
```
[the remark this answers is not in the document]
xernobyl: Done in 28b5293 — `TestTenantSettings_PatchBodyTooLargeRejected` now decodes the response and asserts `code == "content_too_large"`, mirroring `TestTenantSettings_PutBodyTooLargeRejected`. (`decodeData` is just `json.NewDecoder(resp.Body).Decode(v)` with no envelope unwrapping, so it reads the top-level problem `code` directly.) Passing locally against `test_db`.

internal/service/tenant_settings_service.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// notifyChanged tells the registered listener which setting keys a write touched. No-op when
+// no listener is set or nothing changed.
+func (s *TenantSettingsService) notifyChanged(ctx context.Context, tenantID string, changedKeys []string) {
+	if s.listener == nil || len(changedKeys) == 0 {
+		return
+	}
+
+	s.listener.OnSettingsChanged(ctx, tenantID, changedKeys)
```
BhagyaAmarasinghe, line 162: Tenant settings cache can cause missed translations right after settings changes. The API provider reads through CachedTenantSettings ([cmd/api/app.go (line 373)](/private/tmp/hub-pr-89-review/cmd/api/app.go:373)), which serves cached values until TTL expiry ([tenant_settings_cache.go (line 52)](/private/tmp/hub-pr-89-review/internal/service/tenant_settings_cache.go:52)). Settings writes notify only the backfill listener, not the cache ([tenant_settings_service.go (line 149)](/private/tmp/hub-pr-89-review/internal/service/tenant_settings_service.go:149)). If a tenant had “no target” cached, then enables a target, records created after the per-tenant backfill finishes but before cache expiry are skipped by the enqueue gate ([translation_provider.go (line 104)](/private/tmp/hub-pr-89-review/internal/service/translation_provider.go:104))

internal/repository/feedback_records_repository.go
```diff
@@ -158,7 +143,150 @@ func (r *FeedbackRecordsRepository) GetByID(ctx context.Context, id uuid.UUID) (
...
+// a now-stale target). It joins tenant_settings for the per-tenant target. Callers append
+// ordering / keyset / limit clauses.
+const translationBackfillSelectSQL = `
+	SELECT fr.id, ts.settings->>'target_language'
+	FROM feedback_records fr
+	JOIN tenant_settings ts ON ts.tenant_id = fr.tenant_id
+	WHERE fr.field_type = 'text'
+		AND fr.value_text IS NOT NULL AND btrim(fr.value_text) <> ''
+		AND COALESCE(ts.settings->>'target_language', '') <> ''
+		AND fr.translation_lang_key IS DISTINCT FROM ts.settings->>'target_language'`
```
BhagyaAmarasinghe: Translation backfill cannot repair stale translated text after value_text or language changes. Record updates change value_text/language without clearing value_text_translated or translation_lang_key ([feedback_records_repository.go (line 462)](/private/tmp/hub-pr-89-review/internal/repository/feedback_records_repository.go:462)), while the backfill query only selects rows whose translation_lang_key differs from the tenant target ([feedback_records_repository.go (line 220)](/private/tmp/hub-pr-89-review/internal/repository/feedback_records_repository.go:220)). If the update-triggered job is missed, swallowed, or fails finally, stale translated text remains excluded from recovery backfills.

internal/models/tenant_settings.go
```diff
@@ -0,0 +1,38 @@
+package models
+
+// EnrichmentSettings is the typed contents of the tenant_settings.settings JSONB
+// column. Every field is optional and falls back to a documented default applied
+// by the service when absent. This is the forward-compatible surface for
+// tenant-scoped enrichment configuration: add new settings as fields here, no
+// schema migration required.
+type EnrichmentSettings struct {
```
BhagyaAmarasinghe, line 8: AI Review: ENG-1254’s core settings contract is incomplete. The Linear issue requires translation_enabled, target_language, env-default fallback, and explicit disabled-vs-default behavior. This PR only models target_language ([tenant_settings.go (line 8)](/private/tmp/hub-pr-89-review/internal/models/tenant_settings.go:8)), returns empty settings for unset tenants ([tenant_settings_service.go (line 57)](/private/tmp/hub-pr-89-review/internal/service/tenant_settings_service.go:57)), and has no default enabled/language config in TranslationConfig ([config.go (line 133)](/private/tmp/hub-pr-89-review/internal/config/config.go:133)). That means tenants cannot explicitly disable translation while inheriting defaults, and unset tenants never inherit env defaults.

BhagyaAmarasinghe (changes requested):

internal/service/tenant_settings_service.go
xernobyl: Addressed in 2267b2e — the settings cache now evicts the tenant's entry on a settings write (composed with the backfill listener), so a changed or newly enabled target is seen by the enqueue gate immediately instead of after TTL.

internal/repository/feedback_records_repository.go
xernobyl: Addressed in 2267b2e — an update to value_text or language now clears value_text_translated/translation_lang_key, but only on an actual change, so the row falls back to the original and stays recoverable by a backfill.

internal/models/tenant_settings.go
xernobyl: thank you for pointing that, I forgot that we mentioned having a default 😅

xernobyl: I added TRANSLATION_DEFAULT_LANGUAGE env, also updated docs.

BhagyaAmarasinghe: Still lacks translation_enabled and the enabled/disabled env default.
The new commit adds TRANSLATION_DEFAULT_LANGUAGE, but Linear still requires tenant-level translation_enabled plus a global enabled/disabled default. Without that, a tenant cannot explicitly opt out while inheriting default language, so the “disabled vs default” contract is still not implemented.

internal/repository/feedback_records_repository.go
```diff
@@ -158,7 +143,165 @@ func (r *FeedbackRecordsRepository) GetByID(ctx context.Context, id uuid.UUID) (
...
+		tag, err := dbTx.Exec(ctx, `
+			UPDATE feedback_records fr
+			SET value_text_translated = $2, translation_lang_key = $3, updated_at = NOW()
+			FROM (
+				SELECT NULLIF(ts.settings->>'target_language', '') AS stored_target
+				FROM feedback_records r
+				LEFT JOIN tenant_settings ts ON ts.tenant_id = r.tenant_id
+				WHERE r.id = $1
+			) eff
+			WHERE fr.id = $1
```
BhagyaAmarasinghe, line 205: Stale explicit-target jobs can write after a tenant clears its target and falls back to default.
The guard accepts any langKey when eff.stored_target IS NULL. If a tenant had de-DE, an old de-DE job is queued, then the tenant clears target_language to inherit default en-US, that old job can still write de-DE because no stored target exists. The guard needs to compare against the effective target, e.g. COALESCE(stored_target, defaultLang) = langKey.

internal/repository/feedback_records_repository.go
```diff
@@ -158,7 +143,165 @@ func (r *FeedbackRecordsRepository) GetByID(ctx context.Context, id uuid.UUID) (
...
+	const query = translationBackfillSelectSQL + `
+			AND fr.tenant_id = $2
+			AND fr.id > $3
+		ORDER BY fr.id
+		LIMIT $4`
+
+	// Empty default ($1): a settings-change backfill re-translates to the tenant's OWN target
+	// (which it has — changing it is what triggered this). The global default fallback is
+	// applied only by the cross-tenant ListTranslationBackfillTargets.
+	rows, err := r.db.Query(ctx, query, "", tenantID, afterID, limit)
```
BhagyaAmarasinghe: Per-tenant settings-change backfill ignores the default fallback.
ListTranslationBackfillTargetsForTenant always passes "" as the default. If a tenant clears its explicit target, it should inherit TRANSLATION_DEFAULT_LANGUAGE, but the immediate settings-triggered backfill becomes a no-op and existing records stay translated to the old target until a global backfill runs.

internal/repository/feedback_records_repository.go
xernobyl: Addressed in 184eaa1. The guard now compares langKey against the tenant's effective target — `COALESCE(eff.stored_target, NULLIF($defaultLang, '')) = $langKey`, threaded through the service. Your scenario (tenant had de-DE, clears it, falls back to the en-US default, an old de-DE job still in flight) now resolves the effective target to en-US, so the de-DE write no longer matches and is superseded. Covered by a new regression test, `TestFeedbackRecords_SetTranslation_StaleExplicitTargetUnderDefaultIsSuperseded`.

internal/repository/feedback_records_repository.go
xernobyl: Addressed in 184eaa1. `ListTranslationBackfillTargetsForTenant` now takes the deployment default and resolves the effective target with the same `COALESCE(own target_language, default)`. So clearing a tenant's `target_language` no longer no-ops — the settings-change backfill re-translates that tenant's existing records to the default. New regression test: `TestListTranslationBackfillTargetsForTenant_UsesDefaultLanguage`.

internal/models/tenant_settings.go
xernobyl: Two modes here: **without** `TRANSLATION_DEFAULT_LANGUAGE`, translation is pure per-tenant opt-in — a tenant opts in by setting `target_language` and opts out by leaving it unset. **With** a default set, the deployment has deliberately chosen to translate every tenant; a tenant setting `target_language` to the default is just a redundant no-op, and any other value picks a different target — neither disables translation, by design.

If we later need a tenant to opt out *while* a deployment default is set, that's a small fast-follow (a `translation_enabled` flag / sentinel), but I'd rather hold off until a customer actually needs it than reintroduce the "enabled-but-no-target" ambiguity we removed. I've updated ENG-1254 to reflect this design.

BhagyaAmarasinghe (approved):

PR #100, opened 2026-07-06
feat(observability): enrichment round-2 (event correlation, HNSW-degradation gauge) + perf nits

xernobyl:
## What does this PR do?

Round-2 observability + free perf wins for the enrichment framework. Deliberately kept **out of #99** so that PR's review stays focused.

> **#99 has merged — this is now rebased onto `main` and ready for review.** It was stacked on #99 (`refactor/ENG-1613_enrichment-framework`) during review because it touches the classify job args/worker introduced there; those are in `main` now.

Relates to ENG-1644 (observability/perf) and ENG-1613 (the `AGENTS.md` refresh, folded in from the former #101 — not worth a standalone PR).

**Commits (9 total, grouped by theme; each self-contained):**

1. **`perf(embeddings)`** — the nearest/after-cursor search queries selected both `(e.embedding <=> $1) AS distance` *and* `(1 - …) AS score`, evaluating the cosine-distance operator twice per row; drop the score column and derive `score = 1 - distance` in Go. Also build `BuildEmbeddingInput` once on the embedding create path (it was computed twice). No behavior change.
2. **`feat: HNSW iterative-scan gauge`** — when the server rejects `hnsw.iterative_scan` (pgvector < 0.8) the repo latches the fallback and logs *once*, after which recall is silently capped at `ef_search` for the process lifetime. Expose it as `hub_hnsw_iterative_scan_degraded` (0/1), registered alongside the search handler.
3. **`feat: carry event id in job args`** — add `EventID` to all four enrichment job args, populated from `event.ID` at enqueue. Not part of the River unique key (correlation metadata, not a dedupe input); `uuid.Nil` for backfill jobs. Backward-compatible JSON.
4. **`feat: log event id in workers`** *(+ `refactor: complete event_id coverage on the embedding path`, + `fix: correlate the translation short-circuit log line`, + `test: lock event-id propagation and the iterative-scan gauge`)* — emit `event_id` on worker log lines so a job's logs correlate to its originating event. The generic worker binds a `slog.Logger` once (also removing `feedback_record_id` duplicated across ~12 sites); the standalone embedding worker gets its main lines, and the translation worker's same-language short-circuit line is correlated too. The test commit pins event-id propagation through the providers and the gauge's 0/1 behavior.
[...]

PR #101, opened 2026-07-06
docs: refresh AGENTS.md for the enrichment framework

xernobyl:
## What does this PR do?

Refreshes the repo's `AGENTS.md` so agent/contributor guidance matches the tree after the enrichment work. Documentation only — no code changes.

> **Draft, stacked on #100** (`feat/ENG-1644_enrichment-observability`). #99 has merged and #100 is now rebased onto `main`. This documents the enrichment scaffold that lands with #100, so it stays a draft on #100 until that merges, then retargets to `main`. Review #100 first.

**Changes** (single file, `AGENTS.md`):
- **Structure** — adds the `cmd/backfill-*` commands, the four worker enrichment pipelines, and the `internal/` packages that appeared since (`observability`, `llm`/`openai`/`googleai`, `datatypes`, `huberrors`) plus `pkg/cursor` and `pkg/embeddings`.
- **Commands** — corrects `make run` (starts **both** api and worker, was described as api-only); adds the backfill run/build targets.
- **New "Enrichment Framework" section** — the highest-value addition: points at the shared classify scaffold (registry + `EnrichmentClientConfig`, `classifyStructured`/`structuredSpec`, `EnrichmentProvider`, the generic `enrichmentWorker[A, R]`, shared metrics) and the thin per-type recipe, so a new enrichment **reuses the scaffold instead of copy-pasting one** — the exact duplication #99 removed. Notes the read-only/eager-clear + content-supersession invariants and the tenant_id/event_id propagation expectation.
- **Testing** — splits unit (`internal/**`, no DB, `make test-unit`) from integration (`tests/`, needs a pgvector `test_db`, `make tests`); restates the tenant-isolation alternate-path test requirement.
- **Config** — points at the enrichment/OTel env groups documented in `.env.example`.

The **Multi-Tenancy & Data Isolation** section was reviewed and left unchanged — it's already current (covers workers, backfills, River args, embeddings, search, the tenant write-lock, and the lock order the enrichment work relies on).

## How should this be tested?

- Prose/accuracy review. Every referenced `make` target was checked to exist in the `Makefile`, and every referenced path/package to exist in the tree.
- No build/test impact (docs-only); the pre-commit gate (build/lint/unit) passed on commit.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,17 +1,19 @@
 # Repository Guidelines
 
 ## Project Structure & Module Organization
-- `cmd/api/` holds the API server (hub-api): HTTP API, ingestion, record retrieval, tenant/auth; enqueues jobs to River (insert-only). Build/run: `go run ./cmd/api` or `make run`.
-- `cmd/worker/` holds the worker (hub-worker): runs River job workers (webhook delivery, embeddings). No HTTP. Build/run: `go run ./cmd/worker` or `make run-worker`.
-- `internal/` contains core application layers: `api/handlers`, `api/middleware`, `service`, `repository`, `models`, `config`, and `workers`.
-- `pkg/` provides shared utilities (currently `pkg/database`).
+- `cmd/api/` holds the API server (hub-api): HTTP API, ingestion, record retrieval, tenant/auth, semantic search; enqueues jobs to River (insert-only). Build/run: `go run ./cmd/api` or `make run`.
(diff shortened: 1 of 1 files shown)


xernobyl:
Folded into #100 rather than shipped standalone — a separate PR for a single doc file wasn't worth the review overhead. The `AGENTS.md` refresh is commit `690f9f8` there (unchanged); fast-forwarding #100's branch to include it is why GitHub auto-marked this one merged. Review it as part of #100.

PR #105, opened 2026-07-08
feat: add translated taxonomy embeddings

BhagyaAmarasinghe:
## Summary
- add taxonomy-specific embedding jobs keyed separately from raw search embeddings
- embed taxonomy records from `value_text_translated` when present, falling back to `value_text`
- route taxonomy run counts/input through the taxonomy embedding model key and return canonical translated text to the taxonomy service
- enqueue taxonomy re-embedding after translation writes and add a `-taxonomy` backfill mode
- keep semantic search on raw embeddings by excluding taxonomy model rows from the raw HNSW index/search path
- bump the Go toolchain pins to `1.26.5` so Code Quality runs `govulncheck` against the patched stdlib

## Why
Multilingual directories were clustering by language because taxonomy embeddings were built from raw mixed-language text. This keeps normal semantic search embeddings unchanged while giving taxonomy generation a canonicalized embedding space.

## Rollout
- After deploy, run `backfill-embeddings -taxonomy` for existing records before relying on taxonomy generation for pre-existing directories. New/updated/translated records enqueue taxonomy embeddings automatically.
- If `backfill-embeddings -prune-stale-models` is used during rollout, it now preserves both the active raw embedding model and the active taxonomy embedding model.

## Validation
- `go test ./internal/repository ./internal/service ./cmd/backfill-embeddings`
- `DATABASE_URL='postgres://postgres:postgres@localhost:5432/test_db?sslmode=disable' go test ./tests -run 'TestDeleteEmbeddingsForOtherModels|TestTaxonomyAPI_InternalServiceEndpoints|TestTaxonomyNoSource' -timeout 120s`
- `DATABASE_URL='postgres://postgres:postgres@localhost:5432/test_db?sslmode=disable' go test ./tests/... -timeout 120s`
- `make test-unit`
- `make build`
- `PATH='/Users/bhagya/work/bin:'"$PATH" make migrate-validate`
- `GOLANGCI_LINT=/Users/bhagya/work/bin/golangci-lint PATH='/Users/bhagya/work/bin:'"$PATH" make lint-new`
- `GOLANGCI_LINT=/Users/bhagya/work/bin/golangci-lint make fmt`
- `go run golang.org/x/vuln/cmd/govulncheck@v1.3.0 ./...`

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -23,6 +23,7 @@ import (
 	"github.com/formbricks/hub/internal/api/handlers"
 	"github.com/formbricks/hub/internal/api/middleware"
 	"github.com/formbricks/hub/internal/config"
+	"github.com/formbricks/hub/internal/models"
 	"github.com/formbricks/hub/internal/observability"
 	"github.com/formbricks/hub/internal/repository"
 	"github.com/formbricks/hub/internal/service"
@@ -251,6 +252,13 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
 	tenantDataRepo := repository.NewTenantDataRepository(db, cfg.TenantData.PurgeLockTimeout.Duration())
 	embeddingProviderName, embeddingModel := embeddingProviderAndModel(cfg)
 	embeddingModelForDB := embeddingModel
+	taxonomyEmbeddingModel := service.TaxonomyEmbeddingModel(embeddingModelForDB, cfg.Taxonomy.EmbeddingModel)
+
+	taxonomyEmbeddingEnqueueModel := taxonomyEmbeddingModel
+
+	if embeddingProviderName == "" {
+		taxonomyEmbeddingEnqueueModel = ""
+	}
 
 	feedbackRecordsService := service.NewFeedbackRecordsService(
 		feedbackRecordsRepo,
@@ -262,6 +270,7 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
 		cfg.Embedding.MaxAttempts,
 		cfg.Translation.DefaultLanguage,
 	)
+	feedbackRecordsService.SetTaxonomyEmbeddingModel(taxonomyEmbeddingEnqueueModel)
 
 	// The eager-clear (nulling stale enrichment outputs on a value_text edit) fires only on this
 	// API PATCH path, so wire its counter here; the worker/backfill service instances leave it unset.
@@ -428,6 +437,19 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
 			embeddingMetrics,
 		)
 		messageManager.RegisterProvider(embeddingProv)
+
+		if taxonomyEmbeddingEnqueueModel != "" {
+			taxonomyEmbeddingProv := service.NewEmbeddingProviderForInputKind(
+				riverClient,
+				taxonomyEmbeddingEnqueueModel,
+				service.EmbeddingsQueueName,
+				cfg.Embedding.MaxAttempts,
+				docPrefix,
+				embeddingMetrics,
+				models.EmbeddingInputKindTaxonomyTranslated,
+			)
+			messageManager.RegisterProvider(taxonomyEmbeddingProv)
+		}
 	}
 
 	webhooksService := service.NewWebhooksService(webhooksRepo, messageManager, cfg.Webhook.MaxCount, cfg.Webhook.URLBlacklist)
@@ -516,7 +538,7 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
 	taxonomyService := service.NewTaxonomyService(service.NewTaxonomyServiceParams{
 		Repo:                  taxonomyRepo,
 		Starter:               taxonomyStarter,
-		EmbeddingModel:        embeddingModelForDB,
+		EmbeddingModel:        taxonomyEmbeddingModel,
 		MinimumEmbeddingCount: cfg.Taxonomy.MinimumEmbeddedRecords,
 	})
 	taxonomyHandler := handlers.NewTaxonomyHandler(taxonomyService)
--- a/cmd/backfill-embeddings/main.go
+++ b/cmd/backfill-embeddings/main.go
@@ -115,11 +115,12 @@ func run() int {
 	}
 
 	embeddingModelForDB := embeddingModel
+	taxonomyEmbeddingModel := service.TaxonomyEmbeddingModel(embeddingModelForDB, cfg.Taxonomy.EmbeddingModel)
 	targetModel := embeddingModelForDB
 	inputKind := models.EmbeddingInputKindRaw
 
 	if *taxonomyMode {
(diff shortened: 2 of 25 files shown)


cmd/backfill-embeddings/main.go
```diff
@@ -112,11 +115,24 @@ func run() int {
...
 	embeddingsRepo := repository.NewEmbeddingsRepository(db)
 
 	if *pruneStaleModels {
+		if *taxonomyMode {
+			slog.Error("-taxonomy cannot be combined with -prune-stale-models")
+
+			return exitFailure
+		}
+
 		deleted, pruneErr := embeddingsRepo.DeleteEmbeddingsForOtherModels(ctx, embeddingModelForDB, pruneBatchSize)
```
xernobyl: **[P2] This deletes every taxonomy embedding.** `DeleteEmbeddingsForOtherModels` runs `DELETE FROM embeddings WHERE model <> $1` with the raw model, and taxonomy rows are keyed `taxonomy:…` — always ≠ the raw model — so a plain `-prune-stale-models` run (the documented post-migration cleanup) silently empties the taxonomy space. Taxonomy generation then falls below the min-embeddings gate until someone re-runs `-taxonomy`.

The new guard above blocks `-taxonomy` + `-prune-stale-models` together, but that's the harmless combination; the harmful one is prune *alone*. Consider preserving the current `taxonomy:<model>:…` key here, or making prune model-aware.

cmd/api/app.go
```diff
@@ -428,6 +437,19 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
 			embeddingMetrics,
 		)
 		messageManager.RegisterProvider(embeddingProv)
+
+		if taxonomyEmbeddingEnqueueModel != "" {
+			taxonomyEmbeddingProv := service.NewEmbeddingProviderForInputKind(
```
xernobyl, line 442: **[P2] Index growth on the shared HNSW graph.** Registering this second provider means roughly one taxonomy vector per text record on top of the raw one, ~doubling the `embeddings` table. `idx_embeddings` is a single HNSW graph across all `model` values (migrations/005), and semantic search post-filters that same graph with `WHERE e.model = $2 … ORDER BY embedding <=> $1`. This is exactly the pressure the prune code's comment warns about ("every stale vector consumes candidate budget on every search, worsening recall").

Was index growth considered? A partial index or a separate table for non-search embeddings would keep the search path off these vectors.

cmd/api/app.go
```diff
@@ -516,7 +538,7 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
 	taxonomyService := service.NewTaxonomyService(service.NewTaxonomyServiceParams{
 		Repo:                  taxonomyRepo,
 		Starter:               taxonomyStarter,
-		EmbeddingModel:        embeddingModelForDB,
+		EmbeddingModel:        taxonomyEmbeddingModel,
```
xernobyl, line 541: **[P2] Taxonomy is dark for existing data until the `-taxonomy` backfill runs.** The gate now counts embeddings under `taxonomyEmbeddingModel`, which has zero rows at deploy, so every scope returns "at least N embedded text feedback records are required; found 0" (default 20) until the backfill populates them — existing records don't get a taxonomy embedding from the event path unless updated or translated.

Please make sure the deploy plan runs `backfill-embeddings -taxonomy`; it isn't in the PR's validation notes.

internal/service/embedding_provider.go
```diff
@@ -127,6 +144,15 @@ func (p *EmbeddingProvider) PublishEvent(ctx context.Context, event Event) {
 	}
 }
 
+func (p *EmbeddingProvider) hasEmbeddingRelevantChange(changedFields []string) bool {
+	if slices.Contains(changedFields, "value_text") || slices.Contains(changedFields, "field_label") {
+		return true
+	}
+
+	return p.inputKind == models.EmbeddingInputKindTaxonomyTranslated &&
```
xernobyl: **[P3] Unreachable branch.** `UpdateFeedbackRecordRequest.ChangedFields()` never emits `value_text_translated` (models/feedback_records.go), and translation writes go through `SetTranslation`, which publishes no event and enqueues the re-embed directly. So this branch never fires. Harmless, but it reads as if events drive taxonomy re-embedding when they don't — drop it or mark it as forward-compat.

internal/repository/taxonomy_repository.go
```diff
@@ -890,13 +888,12 @@ func (r *TaxonomyRepository) queryRunInputRows(
 				COALESCE(NULLIF(btrim(fr.source_id), ''), ''),
 				fr.field_id,
 				COALESCE(fr.field_label, ''),
-				fr.value_text,
+				COALESCE(NULLIF(btrim(fr.value_text_translated), ''), fr.value_text),
```
xernobyl: **[P3, minor]** Two small inconsistencies: this SELECT falls back to un-`btrim`'d `fr.value_text` while the WHERE uses `btrim` (cosmetic trailing whitespace in the taxonomy input), and `ListFieldOptions` (~line 70) still filters on raw `value_text` only, whereas this query and `CountScopeInput` moved to the translated-or-raw coalesce. Equivalent in practice today (a translation implies non-blank raw text), but a latent divergence between the field picker's counts and the run-start gate.

xernobyl (commented):
Reviewed at head e886878. The core design is sound and well-covered: raw search embeddings are left untouched, taxonomy gets its own `taxonomy:<model>:translated-v1` space, the concurrency guards are correctly threaded through the new `valueTextTranslated` parameter, and the translation-write re-embed path publishes no event so it can't double-fire with the event provider. CI is green and I'm not re-raising CodeRabbit's open note.

A few design/operational findings inline — none break the happy path. The two I'd resolve before merge are the prune data-loss (P2) and a decision on the shared-index growth (P2); the backfill rollout step is important but operational.

xernobyl (approved):
Re-reviewed at f58d675 — thanks for the quick turnaround. All my findings are addressed and the approach is solid:

- Prune now keeps both the raw and taxonomy model keys (with an empty-set guard against a full-table delete).
- The search HNSW index is now partial (`model NOT LIKE 'taxonomy:%'`), with the matching predicate on all four ANN queries, so taxonomy vectors no longer consume search candidate budget.
- Dead `value_text_translated` branch removed; the taxonomy/raw text filters are consistent across `ListFieldOptions`, `CountScopeInput`, and the run-input queries.
- New tests cover the prune keep-set and translated-only records.

A few non-blocking notes for the record:
- The failing Code Quality check is govulncheck GO-2026-5856 (crypto/tls in the go1.26.4 stdlib, fixed in 1.26.5). It's repo-wide, not from this PR — needs a separate toolchain bump.
- Existing data still needs `backfill-embeddings -taxonomy` on rollout before taxonomy runs clear the minimum-embeddings gate.
- Optional hardening: the exclusion keys off the `taxonomy:` prefix, so a custom `TAXONOMY_EMBEDDING_MODEL` override without that prefix would re-enter the search index. Worth enforcing the prefix eventually.

LGTM 👍

PR #109, opened 2026-07-20
 fix(taxonomy): reap stalled runs so taxonomy generation can recover (ENG-1924)

pandeymangg:
## What does this PR do?

A taxonomy run can be orphaned in `pending`/`running` if the taxonomy service crashes mid-run or its terminal callback (`POST /internal/v1/taxonomy/runs/{id}/failed` or the success PUT) is lost. The Hub never transitions the run, so Formbricks Web treats it as "in progress" — it disables Generate and polls `GET /runs/{id}` every 5s **forever**, recoverable only by a manual DB edit.

This adds a periodic Hub-side **reaper** that force-fails runs stalled past a timeout, so polling stops, the failure surfaces, and generation can be retried.

- **`TaxonomyRepository.FailStuckRuns(ctx, olderThan, message, code)`** — a single bulk `UPDATE` marking `pending`/`running` runs older than the cutoff (`COALESCE(started_at, created_at)`) as `failed`. Idempotent and race-safe: the `status IN ('pending','running')` filter skips runs that finished on their own between sweeps.
- **`runTaxonomyRunReaper`** — a ticker poller mirroring the existing `runRiverQueueDepthPoller`, launched in the API process only when the taxonomy service is configured. Cross-tenant, idempotent, safe under multiple replicas.
- **Config** — `TAXONOMY_STUCK_RUN_TIMEOUT_SECONDS` (default 1800) and `TAXONOMY_REAPER_INTERVAL_SECONDS` (default 60), via the existing `DurationSec` pattern.

**Decisions flagged for review:**
- Reaped runs are failed with `error_code = internal_error` and the raw message *"taxonomy run timed out without completing"* (operator-facing; Formbricks Web maps the code to a localized message — ENG-1917). Alternative: `service_unavailable` (retry-friendlier). Easy to flip.
- ⚠️ The default timeout (30 min) **must exceed the longest legitimate run duration**, or healthy long-running generations get reaped. Tune to the real max + margin.

No schema/migration changes (uses existing columns). No API changes.

Closes ENG-1924 · Follow-up to ENG-1213 · pairs with the Web-side wedge fix ENG-1917.

## How should this be tested?

- **Automated:** `TestTaxonomyRepository_FailStuckRuns` (in `tests/`) creates a stuck `running` run (aged `started_at`), a stuck `pending` run (aged `created_at`), and a fresh run, then asserts the reaper fails the two stuck ones and leaves the fresh one. Runs via `make tests` (needs `test_db`) — executed in CI.
- **Manual (end-to-end):** with the taxonomy service configured, start a run and stop the taxonomy service mid-run so it's orphaned in `running`; within `TAXONOMY_STUCK_RUN_TIMEOUT_SECONDS + TAXONOMY_REAPER_INTERVAL_SECONDS` the reaper marks it `failed`, the UI stops polling, and Generate re-enables.
- Verified locally: `go build ./...`, `make fmt`, `make lint` (0 issues), `go vet ./tests/`.

### Required

- [ ] Read [Repository Guidelines](https://github.com/formbricks/hub/blob/main/AGENTS.md)
- [ ] Self-reviewed my own code
- [ ] Commented on my code in hard-to-understand bits
- [ ] Ran `make build`
- [ ] Ran `make tests` (integration tests in `tests/`)
- [ ] Merged the latest changes from main onto my branch with `git pull origin main`
- [ ] If database schema changed: added migration in `migrations/` with goose annotations and ran `make migrate-validate` — _N/A, no schema change_

### Appreciated

- [ ] If API changed: added or updated OpenAPI spec and ran contract tests (`make tests` or API contract workflow) — _N/A, no API change_
- [ ] If API behavior changed: added request/response examples or Swagger UI screenshots to this PR — _N/A_

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -215,6 +215,11 @@ type TaxonomyConfig struct {
 	HubInternalAPIToken    string `env:"HUB_INTERNAL_API_TOKEN"`
 	EmbeddingModel         string `env:"TAXONOMY_EMBEDDING_MODEL"`
 	MinimumEmbeddedRecords int    `env:"TAXONOMY_MIN_EMBEDDED_RECORDS" env-default:"20"`
+	// StuckRunTimeout is how long a run may sit in pending/running before the reaper force-fails it;
+	// it must exceed the longest legitimate run so healthy long-running generations are not reaped.
+	StuckRunTimeout DurationSec `env:"TAXONOMY_STUCK_RUN_TIMEOUT_SECONDS" env-default:"1800"`
+	// ReaperInterval is how often the reaper sweeps for stuck runs.
+	ReaperInterval DurationSec `env:"TAXONOMY_REAPER_INTERVAL_SECONDS" env-default:"60"`
 }
 
 // TenantDataConfig holds tenant data purge settings.
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -215,8 +215,11 @@ type TaxonomyConfig struct {
 	HubInternalAPIToken    string `env:"HUB_INTERNAL_API_TOKEN"`
 	EmbeddingModel         string `env:"TAXONOMY_EMBEDDING_MODEL"`
 	MinimumEmbeddedRecords int    `env:"TAXONOMY_MIN_EMBEDDED_RECORDS" env-default:"20"`
-	// StuckRunTimeout is how long a run may sit in pending/running before the reaper force-fails it;
-	// it must exceed the longest legitimate run so healthy long-running generations are not reaped.
+	// StuckRunTimeout is the maximum time a pending/running run may go without its updated_at being
+	// bumped (via the internal heartbeat endpoint) before the reaper force-fails it. Once the taxonomy
+	// service heartbeats during generation this can be tuned down to a small multiple of the heartbeat
+	// interval; until then updated_at only advances on state changes, so keep it above the longest
+	// legitimate run so healthy long-running generations are not reaped.
 	StuckRunTimeout DurationSec `env:"TAXONOMY_STUCK_RUN_TIMEOUT_SECONDS" env-default:"1800"`
 	// ReaperInterval is how often the reaper sweeps for stuck runs.
 	ReaperInterval DurationSec `env:"TAXONOMY_REAPER_INTERVAL_SECONDS" env-default:"60"`
--- a/internal/repository/taxonomy_repository.go
+++ b/internal/repository/taxonomy_repository.go
@@ -8,6 +8,7 @@ import (
 	"log/slog"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5"
@@ -326,6 +327,33 @@ func (r *TaxonomyRepository) MarkRunFailed(
 	}
 
 	return run, nil
+}
+
+// FailStuckRuns marks taxonomy runs stuck in a non-terminal state (pending/running) past olderThan
+// as failed. Runs are orphaned when the taxonomy service crashes mid-run or its terminal callback is
+// lost; without this sweep they are polled forever in the UI and block regeneration. The status
+// filter keeps it idempotent and race-safe — a run that finishes between sweeps no longer matches.
+// Returns the number of runs failed.
+func (r *TaxonomyRepository) FailStuckRuns(
+	ctx context.Context,
+	olderThan time.Duration,
+	message string,
+	errorCode models.TaxonomyRunFailureCode,
+) (int64, error) {
+	cutoff := time.Now().Add(-olderThan)
+
+	tag, err := r.db.Exec(ctx, `
+		UPDATE taxonomy_runs
+		SET status = 'failed', error = $1, error_code = $2, finished_at = NOW(), updated_at = NOW()
+		WHERE status IN ('pending', 'running')
+		  AND COALESCE(started_at, created_at) < $3`,
+		message, nullableFailureCode(errorCode), cutoff,
+	)
+	if err != nil {
+		return 0, fmt.Errorf("fail stuck taxonomy runs: %w", err)
+	}
+
+	return tag.RowsAffected(), nil
 }
 
 // GetRunForInternalService returns run metadata for internal taxonomy service-token workflows.
--- a/internal/repository/taxonomy_repository.go
+++ b/internal/repository/taxonomy_repository.go
@@ -430,9 +430,16 @@ func (r *TaxonomyRepository) FailStuckRuns(
 			continue
 		}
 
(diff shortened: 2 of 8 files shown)


internal/repository/taxonomy_repository.go
```diff
@@ -328,6 +329,33 @@ func (r *TaxonomyRepository) MarkRunFailed(
...
+	message string,
+	errorCode models.TaxonomyRunFailureCode,
+) (int64, error) {
+	cutoff := time.Now().Add(-olderThan)
+
+	tag, err := r.db.Exec(ctx, `
+		UPDATE taxonomy_runs
+		SET status = 'failed', error = $1, error_code = $2, finished_at = NOW(), updated_at = NOW()
+		WHERE status IN ('pending', 'running')
+		  AND COALESCE(started_at, created_at) < $3`,
```
BhagyaAmarasinghe: `started_at` is never refreshed, so this condition treats every run exceeding the timeout as abandoned—even if it is still processing normally. A healthy phased taxonomy generation can exceed the 30-minute default, after which its successful callback will be rejected because the run is already failed. Please use a heartbeat/lease timestamp, or establish and enforce a proven maximum runtime across both services.

internal/repository/taxonomy_repository.go
```diff
@@ -328,6 +329,33 @@ func (r *TaxonomyRepository) MarkRunFailed(
...
+// Returns the number of runs failed.
+func (r *TaxonomyRepository) FailStuckRuns(
+	ctx context.Context,
+	olderThan time.Duration,
+	message string,
+	errorCode models.TaxonomyRunFailureCode,
+) (int64, error) {
+	cutoff := time.Now().Add(-olderThan)
+
+	tag, err := r.db.Exec(ctx, `
```
BhagyaAmarasinghe: This direct cross-tenant `UPDATE` bypasses `withTenantWritePoolTx`, violating the repository invariant that every tenant-owned mutation participates in the shared advisory lock. It can race with a tenant-data purge and make the purge wait on row locks. Please batch or scope candidates by tenant and perform the updates through the tenant write-lock helper, with a purge-race regression test.

internal/config/config.go
```diff
@@ -215,6 +215,11 @@ type TaxonomyConfig struct {
 	HubInternalAPIToken    string `env:"HUB_INTERNAL_API_TOKEN"`
 	EmbeddingModel         string `env:"TAXONOMY_EMBEDDING_MODEL"`
 	MinimumEmbeddedRecords int    `env:"TAXONOMY_MIN_EMBEDDED_RECORDS" env-default:"20"`
+	// StuckRunTimeout is how long a run may sit in pending/running before the reaper force-fails it;
+	// it must exceed the longest legitimate run so healthy long-running generations are not reaped.
+	StuckRunTimeout DurationSec `env:"TAXONOMY_STUCK_RUN_TIMEOUT_SECONDS" env-default:"1800"`
+	// ReaperInterval is how often the reaper sweeps for stuck runs.
+	ReaperInterval DurationSec `env:"TAXONOMY_REAPER_INTERVAL_SECONDS" env-default:"60"`
```
BhagyaAmarasinghe, line 225: These settings are critical because an undersized timeout can fail legitimate runs, but neither variable is documented in .env.example. Please add both variables with their units, defaults, and a warning that the timeout must exceed the longest legitimate taxonomy run.

BhagyaAmarasinghe (changes requested):
Some comments including AI review

xernobyl (commented):
Two low-severity notes, neither blocking — the change is correct and CI is green. Security pass (gitleaks + manual injection/PII/egress check) came back clean.

internal/repository/taxonomy_repository.go
xernobyl: **[P3] Bypasses the tenant write-lock convention**

Every other tenant-owned mutation goes through `withTenantWritePoolTx` to take the shared advisory lock that serializes writes against a tenant data purge (see `tenant_write_lock.go`). This sweep writes on the raw pool, so it can run concurrently with a purge.

I don't think it's exploitable today: it's UPDATE-only on status/error/timestamp columns, never inserts, and never touches `id`/`tenant_id` — so the purge's `run_id IN (SELECT id FROM taxonomy_runs WHERE tenant_id = $1)` child deletes and the final `DELETE FROM taxonomy_runs` see the same rows regardless, and there's no way to orphan children or resurrect purged data. `runRiverQueueDepthPoller` already accesses the pool cross-tenant the same way.

The risk is future-proofing: if this method ever grows an INSERT/DELETE, the bypass silently becomes unsafe. Could you add a one-line comment noting it's intentionally exempt from the write-lock convention (cross-tenant janitor, UPDATE-only)? Not a blocker.

internal/repository/taxonomy_repository.go
xernobyl: **[P3] Cross-tenant predicate isn't index-supported → recurring seq scan**

All three `taxonomy_runs` indexes are `tenant_id`-prefixed, and `COALESCE(started_at, created_at)` is an unindexed expression. This sweep has no `tenant_id` filter, so each pass (every `TAXONOMY_REAPER_INTERVAL_SECONDS`, per replica, plus once per pod on startup) seq-scans `taxonomy_runs`.

Negligible now, but it grows with the table since runs are only removed by purge. If we want to keep it cheap long-term, a partial index would make the lookup O(stuck rows):

```sql
CREATE INDEX idx_taxonomy_runs_stuck
  ON taxonomy_runs ((COALESCE(started_at, created_at)))
  WHERE status IN ('pending', 'running');
```

Since the PR intentionally avoids a migration, fine to leave as a follow-up. Not a blocker.

internal/repository/taxonomy_repository.go
pandeymangg: resolved now 🙏

internal/repository/taxonomy_repository.go
pandeymangg: Good call, and agreed on leaving it as a follow-up. One update since you commented: the predicate changed with the heartbeat work — the sweep now filters on updated_at < cutoff (the liveness signal the heartbeat endpoint bumps) instead of COALESCE(started_at, created_at). The seq-scan observation still holds though: no tenant_id filter, and updated_at isn't covered by the tenant_id-prefixed indexes.

When we do pick it up, the partial index should target the new column:
```
CREATE INDEX idx_taxonomy_runs_stuck
    ON taxonomy_runs (updated_at)
    WHERE status IN ('pending', 'running');
```
Keeping it out of this PR to stay migration-free as intended — I'll file a follow-up to track it.

xernobyl (approved):
Thank you for the changes. LGTM.

BhagyaAmarasinghe (approved):

PR #111, opened 2026-07-24
feat: expose per-tenant enrichment status and backlog metric (ENG-1670)

xernobyl:
## What does this PR do?

Gives the product a way to surface enrichment progress ("in progress" / "done") for a feedback directory, the way we already do for embeddings.

New endpoint: `GET /v1/enrichment-status?tenant_id=<id>` returns, per record-level enrichment (translation, sentiment, emotions):

```json
{
  "tenant_id": "…",
  "translation": { "enabled": true, "eligible": 500, "done": 480 },
  "sentiment":   { "enabled": true, "eligible": 500, "done": 500 },
  "emotions":    { "enabled": false, "eligible": 0, "done": 0 }
}
```

- `eligible` = feedback records that qualify for the enrichment (open-text with content, and the enrichment is on for the tenant — a resolvable target language for translation, the per-directory switch for sentiment/emotions).
- `done` = how many have been enriched. The UI derives "in progress" as `eligible - done` (same as the embeddings banner).
- `enabled` folds in the deployment-level provider/model gate; when it's false the counts are zero.

**Design note — this is data-derived, not queue-derived.** The counts come from `feedback_records` (mirroring how `/v1/taxonomy/fields` reports `record_count`/`embedding_count`), not from the River queue. That choice is deliberate: it gives the progress-bar denominator the queue can't, and — importantly — it needs **no changes to the ingest/enqueue path**, so a burst of a few thousand incoming records is unaffected (no new index on the insert path, no per-job tagging). The per-tenant query is served by the existing `idx_feedback_records_tenant_field_type`.

### Also included

**A schema change — `emotions_classified_at` (migration 020).** Review surfaced that emotions could never reach "done": a successful classification that detects *no* emotion is stored as NULL (the 015 CHECK rejects the empty array), which is the same value that means "never classified". Those records would have counted as pending forever and the backlog gauge could never drain. The new column records *completion* independently of the labels found:

- `SetEmotions` stamps it (an empty result is still a completed classification); a new `ClearEmotions` clears both, for the content-is-gone path.
- The eager-clear on a `value_text` edit drops the marker with the labels, so an edited record correctly returns to pending.
- `done` is `(emotions_classified_at IS NOT NULL OR emotions IS NOT NULL)` — the second arm covers rows enriched before the column existed, which is why the migration needs **no bulk UPDATE** of the primary write table.
- Bonus: the classify backfill now skips already-classified rows instead of re-sending every classified-empty record to the provider on every run.

The marker is processing state and is deliberately kept out of `feedbackRecordColumns` / the API model, so there is **no response-shape, OpenAPI or SDK change**. Sentiment and translation are unaffected — their successful results are single-valued and always non-NULL (verified); emotions is the only enrichment whose success can legitimately be "nothing".

**Backlog metric.** `hub_enrichment_pending_records{enrichment}` — an OpenTelemetry gauge of the aggregate (cross-tenant) eligible-but-unenriched backlog, a durable completeness signal complementing the transient River queue-depth gauge. It only runs when metrics are enabled, is bounded by a 30s statement timeout, and polls every 5 minutes (a full-table aggregate over a high-write table, so we keep it infrequent). Labeled by enrichment type only — never `tenant_id` — to keep cardinality bounded.

Production runs several API replicas per region, so the poller is **leader-elected**: a session-scoped advisory lock held on a dedicated connection for the process lifetime means exactly one replica scans and exports the (global by definition) series. Leadership is deliberately sticky rather than per-scan — replicas tick on independent schedules, so a lock held only for the duration of a scan would almost never collide, reducing neither the DB load nor the duplicate series, while occasionally blanking a replica's series.

Failed refreshes increment `hub_enrichment_backlog_poll_errors_total` and escalate warn → error after three consecutive misses, because a failing poll otherwise leaves the gauge frozen at its last value — indistinguishable from a healthy steady backlog on a dashboard. A scan cancelled by shutdown is not counted (it would fire that alert on every rolling deploy).

**Dependency bump.** `golang.org/x/text` 0.37.0 → 0.39.0 for GO-2026-5970, which was failing the govulncheck gate on unrelated pre-existing call paths.

Out of scope (follow-ups): the Formbricks UI that consumes this, and a per-tenant `failed` count (permanent failures are already observable globally via `hub_<type>_outcomes_total{status="failed_final"}`).

Linear: https://linear.app/formbricks/issue/ENG-1670

## How should this be tested?

**Automated**

- `make tests` — integration tests in `tests/enrichment_status_test.go` cover the SQL against Postgres: eligibility (text + content only, whitespace-only and non-text excluded), translation done vs. stale (`translation_lang_key` = effective target), the sentiment **and** emotions per-tenant switches, the default-language fallback, strict per-tenant isolation, and the aggregate delta.
- Emotion completion specifically: a classified-empty record counts as **done** (not pending forever), a row enriched before the marker existed still counts as done, an unprocessed row stays pending, and editing the text returns a record to pending.
- Leader election: one replica wins and returns the true aggregate, a second is denied (a normal skip, not an error) and publishes nothing, leadership is sticky across polls, and `Close` hands over promptly.
- `go test ./internal/...` — service gating/zeroing, handler 400/503/200, the gauge and the poll-error counter, and that empty content takes the *clear* path rather than recording a classification.
- `make migrate-validate` for migration 020.

**Manual (live progression)**

Bring up `hub-api` + `hub-worker` against a DB with translation + sentiment configured, set a target language for a tenant, create a few open-text records, then poll the endpoint. Config used: `DATABASE_URL`, `API_KEY`, `TRANSLATION_PROVIDER/MODEL/_API_KEY`, `SENTIMENT_PROVIDER/MODEL/_API_KEY`, `TRANSLATION_DEFAULT_LANGUAGE`.

```
curl -H "Authorization: Bearer $API_KEY" "$HUB/v1/enrichment-status?tenant_id=demo"
```

Created 3 text records, then polled:

```jsonc
// T0 — immediately after creating
{"tenant_id":"demo-eng1670","translation":{"enabled":true,"eligible":3,"done":1},"sentiment":{"enabled":true,"eligible":3,"done":0},"emotions":{"enabled":false,"eligible":0,"done":0}}

// T1 — ~12s later, once the workers drained the queue
{"tenant_id":"demo-eng1670","translation":{"enabled":true,"eligible":3,"done":3},"sentiment":{"enabled":true,"eligible":3,"done":3},"emotions":{"enabled":false,"eligible":0,"done":0}}
```

`done` climbs from the partial state to `eligible` as enrichment completes — that's the "in progress → done" signal the UI needs. (`emotions` is `enabled:false` here only because that provider wasn't configured in the test env.) Also verified: `401` without the key, `400` when `tenant_id` is missing, and isolation (an unknown tenant returns zero counts).

### Required

- [x] Ran `make build`
- [x] If database schema changed: added migration in `migrations/020_add_emotions_classified_at.sql` with goose annotations and ran `make migrate-validate`. Nullable `ADD COLUMN` with no default, so it is metadata-only (instant, no rewrite, no long lock) and re-runnable; no data backfill by design (see the note above)

### Appreciated

- [x] If API changed: added or updated OpenAPI spec and ran contract tests (`make lint-openapi`)
- [ ] Updated docs in `docs/` — the Hub API reference regenerates from `openapi.yaml` via Stainless; the hand-authored `hub-api-docs` pages (metrics reference + core-concepts) are a separate follow-up


xernobyl:
@coderabbitai review

cmd/api/app.go
```diff
@@ -692,6 +715,15 @@ func (a *App) Run(ctx context.Context) error {
 		go runRiverQueueDepthPoller(ctx, a.db, a.metrics.Events)
 	}
 
+	if a.metrics != nil && a.metrics.EnrichmentBacklog != nil {
+		go runEnrichmentBacklogPoller(ctx, a.db, a.metrics.EnrichmentBacklog, enrichmentBacklogPollConfig{
+			defaultLang:           a.cfg.Translation.DefaultLanguage,
```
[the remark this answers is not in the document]
xernobyl: Done in c23c285 — trimmed the poller's `defaultLang` for symmetry with `NewEnrichmentStatusService`.

Worth noting the two paths couldn't actually diverge: `config.Load` already canonicalizes `TRANSLATION_DEFAULT_LANGUAGE` (trim + `language.Parse`) at startup, so the endpoint and the gauge always saw the same target. Added the trim anyway as defensive symmetry so the two construction sites read consistently.

internal/repository/enrichment_status_repository.go
```diff
@@ -0,0 +1,128 @@
...
+// idx_feedback_records_tenant_field_type (tenant_id, field_type): only the tenant's text rows are
+// visited, so cost scales with the tenant's text-record count, not the whole table.
+const countEnrichmentStatusSQL = `SELECT ` + enrichmentCountSelect + enrichmentCountFrom + `
+	WHERE fr.tenant_id = $2`
+
+// countEnrichmentBacklogAggregateSQL is the same SELECT without the tenant filter: it sums
+// eligible/done per enrichment across ALL tenants (for the observability gauge). $1 = deployment
+// default target language. The per-tenant enable gates still apply, so a tenant that switched an
+// enrichment off, or has no resolvable target, never inflates the backlog.
+const countEnrichmentBacklogAggregateSQL = `SELECT ` + enrichmentCountSelect + enrichmentCountFrom
```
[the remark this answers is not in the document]
xernobyl: Good catch — fixed in c23c285. Hoisted `fr.field_type = 'text'` into the outer `WHERE` of both `countEnrichmentStatusSQL` and `countEnrichmentBacklogAggregateSQL`.

It's behavior-preserving: every counted column already implies `field_type = 'text'` via `enrichmentEligibleText`, so no count changes (the integration test that seeds a non-text row still asserts it's excluded). And `EXPLAIN` now shows an index scan over the `(tenant_id, …)` path instead of visiting every row and discarding non-text inside the FILTERs. Updated the doc comment to match.

openapi.yaml
```diff
@@ -1380,6 +1382,56 @@ paths:
...
+                    content:
+                        application/problem+json:
+                            schema:
+                                $ref: '#/components/schemas/ErrorModel'
+                default:
+                    description: Error
+                    content:
+                        application/problem+json:
+                            schema:
+                                $ref: '#/components/schemas/ErrorModel'
```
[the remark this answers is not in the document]
xernobyl: Added in c23c285.

One wording tweak vs. the suggestion: this endpoint isn't feature-gated (the service is always wired in `NewApp`), so the 503 is a defensive guard rather than a deployment toggle. I documented it as "the enrichment status service is not available (code `service_unavailable`)" rather than "not available in this deployment / retry once configured", to avoid implying a switch that doesn't exist. Structure matches the other endpoints' 503 + `ErrorModel`.

internal/repository/enrichment_status_repository.go
```diff
@@ -0,0 +1,137 @@
...
+const countEnrichmentStatusSQL = `SELECT ` + enrichmentCountSelect + enrichmentCountFrom + `
+	WHERE fr.tenant_id = $2 AND fr.field_type = 'text'`
+
+// countEnrichmentBacklogAggregateSQL is the same SELECT without the tenant filter: it sums
+// eligible/done per enrichment across ALL tenants (for the observability gauge). $1 = deployment
+// default target language. The field_type = 'text' predicate narrows the scan to text rows; the
+// per-tenant enable gates still apply, so a tenant that switched an enrichment off, or has no
+// resolvable target, never inflates the backlog.
+const countEnrichmentBacklogAggregateSQL = `SELECT ` + enrichmentCountSelect + enrichmentCountFrom + `
+	WHERE fr.field_type = 'text'`
```
Dhruwang, line 124: No `tenant_id` here, so `idx_feedback_records_tenant_field_type` won't apply — `tenant_id` is the leading column and `field_type = 'text'` alone can't use it. That's a full scan of `feedback_records` every 5 min. Did EXPLAIN confirm the plan? Probably want a partial index on `field_type WHERE field_type = 'text'`.

cmd/api/app.go
```diff
@@ -730,6 +765,66 @@ var riverDepthQueues = []string{
...
+	ticker := time.NewTicker(enrichmentBacklogInterval)
+	defer ticker.Stop()
+
+	update := func() {
+		queryCtx, cancel := context.WithTimeout(ctx, enrichmentBacklogQueryTimeout)
+		defer cancel()
+
+		counts, err := repo.CountEnrichmentBacklogAggregate(queryCtx, cfg.defaultLang)
+		if err != nil {
+			slog.WarnContext(ctx, "enrichment backlog poll failed", "error", err)
```
Dhruwang: If this keeps failing we just log and hold the last gauge value indefinitely — from a dashboard it looks like a healthy steady backlog. Can we drop to Error (or bump a poll-error counter) after a few consecutive misses so it's alertable?

internal/repository/enrichment_status_repository.go
```diff
@@ -0,0 +1,137 @@
...
+
+// enrichmentCountSelect is the shared six-column SELECT list — {sentiment, emotions, translation} ×
+// {eligible, done} — used by both the per-tenant and aggregate queries so the predicates can't
+// drift between them. Column order must match the EnrichmentStatusCounts scan order below. All
+// fragments are static constants (never user input).
+const enrichmentCountSelect = `
+		COUNT(*) FILTER (WHERE ` + enrichmentEligibleText + ` AND ` + enrichmentSentimentOn + `),
+		COUNT(*) FILTER (WHERE ` + enrichmentEligibleText + ` AND ` + enrichmentSentimentOn + ` AND fr.sentiment IS NOT NULL),
+		COUNT(*) FILTER (WHERE ` + enrichmentEligibleText + ` AND ` + enrichmentEmotionsOn + `),
+		COUNT(*) FILTER (WHERE ` + enrichmentEligibleText + ` AND ` + enrichmentEmotionsOn + ` AND fr.emotions IS NOT NULL),
```
BhagyaAmarasinghe: `emotions IS NOT NULL` cannot represent completion. A successful classification may validly return zero emotions, and the worker deliberately persists that result as NULL. Such records remain permanently “in progress” and inflate the backlog gauge forever. Add a distinct completion marker/content hash, or persist an empty successful result distinctly from “not processed.”

cmd/api/app.go
```diff
@@ -692,6 +716,17 @@ func (a *App) Run(ctx context.Context) error {
 		go runRiverQueueDepthPoller(ctx, a.db, a.metrics.Events)
 	}
 
+	if a.metrics != nil && a.metrics.EnrichmentBacklog != nil {
```
BhagyaAmarasinghe, line 722: The deployment-wide backlog poller runs once per API replica. Production runs 2–4 replicas in both EU and KSA , so every replica executes the same aggregate scan and emits the same global gauge. This multiplies DB load and makes metric aggregation duplicate or collide. Use leader election, an advisory lock, or a singleton collector.

BhagyaAmarasinghe (changes requested):

internal/repository/enrichment_status_repository.go
xernobyl: You're right, and thanks — I'd only EXPLAINed the per-tenant query, not this one. Confirmed: the aggregate plans a `Seq Scan on feedback_records`, exactly as you said — `tenant_id` is the index's leading column and this query has no tenant predicate, so it can never use it. The comment was overclaiming and is fixed in c23c285/ef4af84.

I went the other way on the partial index though. The aggregate has to read every text row regardless, so an index scan covering most of the table isn't cheaper, and this PR is deliberately protecting the ingest path (bursts of a few thousand records) from extra write amplification. Instead the scan cost is bounded three ways: it runs every 5 min rather than 60s, under a 30s statement timeout, and now on exactly one replica via leader election (your and Bhagya's other thread). Happy to add the index if you'd still prefer it — but I'd want to measure the plan on a large tenant first rather than pay for it on every insert.

cmd/api/app.go
xernobyl: Good catch — fixed in b8d3e8b. Failures now increment a dedicated counter, `hub_enrichment_backlog_poll_errors_total`, so the stale-gauge case is alertable rather than invisible, and the log escalates warn → error after three consecutive misses (a single blip is noise; a run of them means the gauge is frozen and lying).

One extra thing that fell out of a later review: a scan cancelled by shutdown was also landing on that counter, which would have fired the very alert on every rolling deploy. That's now excluded (ef4af84).

internal/repository/enrichment_status_repository.go
xernobyl: Confirmed and fixed in ef4af84 — you were right, and it was worse than a rounding error: neutral/factual feedback plausibly classifies to no emotion fairly often, so a large share of records would have sat "in progress" forever and the backlog gauge could never have drained.

Went with your first suggestion, a distinct completion marker: `emotions_classified_at` (migration 020).

- `SetEmotions` stamps it — an empty result is still a completed classification — and a new `ClearEmotions` clears both, for the content-is-gone path (the worker previously used one call for both, which is what made the two states indistinguishable).
- The eager-clear on a `value_text` edit drops the marker with the labels, so an edited record correctly returns to pending.
- "done" is `(emotions_classified_at IS NOT NULL OR emotions IS NOT NULL)`. The second arm covers rows enriched before the column existed, which is why the migration needs no bulk UPDATE of `feedback_records` — only historical classified-empty rows stay unresolved, and the classify backfill re-processes those once and stamps them.
- Bonus from having the marker: the classify backfill now skips already-classified rows instead of re-sending every classified-empty record to the provider on every run.

Kept it out of `feedbackRecordColumns`/the API model, so there's no response-shape, OpenAPI or SDK change — it's processing state, not data.

I also checked whether the same flaw exists elsewhere: it doesn't. Sentiment always produces exactly one label+score on success, and translation always produces text and stamps the target lang key (same-language input copies through). Emotions is the only enrichment whose successful result is a set that can legitimately be empty, which is precisely why it hit this.

cmd/api/app.go
xernobyl: Fixed in ef4af84 — and worth flagging that my first attempt at this (b8d3e8b) was wrong, so please look at the current shape rather than that commit.

That first version took a transaction-scoped `pg_try_advisory_xact_lock` around each scan. A follow-up review caught that it doesn't do what it claims: the lock is only held for the couple of seconds a scan takes, and replicas tick on independent, unsynchronized schedules (each ticker starts at its own boot), so they'd essentially never collide — neither the multiplied DB load nor the duplicate series was actually prevented. Worse, on the rare collision the loser skipped publishing, and a replica that lost on its *first* poll exported no series at all until the next tick — likeliest exactly during a rolling deploy.

It's now sticky leadership, which is what you asked for: a session-scoped `pg_try_advisory_lock` held on a dedicated connection for the process lifetime. Exactly one replica scans and exports the (global by definition) series, followers never pin a connection, the lock is explicitly released before the connection goes back to the pool (a session lock survives `Release()`), a failed scan drops leadership so another replica takes over, and `Close` hands over promptly on shutdown instead of waiting for the session to be reaped. Covered by a DB-backed test that asserts stickiness and handover, not just mutual exclusion.

One consequence worth naming: the gauge now exists on one replica rather than all of them, so there's a gap of up to one interval (5 min) if the leader pod dies. That seemed the right trade against N× scans and N duplicate series — shout if you'd rather have every replica publish and aggregate with `max`.

BhagyaAmarasinghe (approved):

PR #117, opened 2026-07-31
test: strengthen taxonomy API contracts (ENG-1202)

BhagyaAmarasinghe:
## Summary

- centralize public and internal taxonomy route registration so production and integration tests cannot drift
- verify the public route inventory, operation IDs, and response statuses bidirectionally against `openapi.yaml` while keeping internal routes private
- extend taxonomy HTTP contracts for RFC 9457 errors, heartbeat, service availability, missing/invalid runs, invalid transitions, and tenant isolation

Linear: [ENG-1202](https://linear.app/formbricks/issue/ENG-1202/add-hub-api-contract-tests)

## Validation

- `go test -run "^TestTaxonomyAPI" -count=1 ./tests`
- `go test ./tests/... -count=1 -timeout 120s` against a migrated pgvector Postgres database
- `make test-unit`
- `make build`
- `make lint-new`
- `make lint-openapi`
- downstream taxonomy client: `uv run pytest tests/test_hub_client.py tests/test_start_run.py tests/test_taxonomy_domain.py -q` (`34 passed`)

### DIFF
--- a/internal/api/routes/taxonomy.go
+++ b/internal/api/routes/taxonomy.go
@@ -7,43 +7,18 @@ import (
 	"github.com/formbricks/hub/internal/api/handlers"
 )
 
-type publicTaxonomyHandlerID uint8
-
-const (
-	publicListFields publicTaxonomyHandlerID = iota
-	publicCreateRun
-	publicListRuns
-	publicGetActiveTree
-	publicGetRun
-	publicGetTree
-	publicRecordCounts
-	publicRenameNode
-	publicRemoveNode
-	publicListNodeRecords
-)
-
-type internalTaxonomyHandlerID uint8
-
-const (
-	internalAuthCheck internalTaxonomyHandlerID = iota
-	internalGetRunInput
-	internalCompleteRun
-	internalFailRun
-	internalHeartbeat
-)
-
 type publicTaxonomyRoute struct {
 	method           string
 	path             string
 	operationID      string
 	responseStatuses []string
-	handlerID        publicTaxonomyHandlerID
+	handler          func(*handlers.TaxonomyHandler, http.ResponseWriter, *http.Request)
 }
 
 type internalTaxonomyRoute struct {
-	method    string
-	path      string
-	handlerID internalTaxonomyHandlerID
+	method  string
+	path    string
+	handler func(*handlers.TaxonomyInternalHandler, http.ResponseWriter, *http.Request)
 }
 
 var publicTaxonomyRoutes = []publicTaxonomyRoute{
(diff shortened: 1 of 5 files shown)


tests/taxonomy_support_test.go
```diff
@@ -125,36 +153,28 @@ func setupTaxonomyAPIServer(t *testing.T) *taxonomyTestServer {
...
 	// Internal taxonomy routes (separate internal-service token auth).
 	internal := http.NewServeMux()
-	internal.HandleFunc("GET /internal/v1/taxonomy/auth-check", taxonomyInternalHandler.AuthCheck)
-	internal.HandleFunc("GET /internal/v1/taxonomy/runs/{run_id}/input", taxonomyInternalHandler.GetRunInput)
-	internal.HandleFunc("PUT /internal/v1/taxonomy/runs/{run_id}/result", taxonomyInternalHandler.CompleteRun)
-	internal.HandleFunc("POST /internal/v1/taxonomy/runs/{run_id}/failed", taxonomyInternalHandler.FailRun)
+	routes.RegisterInternalTaxonomy(internal, taxonomyInternalHandler)
 	internalWithAuth := middleware.Auth(testInternalToken)(internal)
 
 	mux := http.NewServeMux()
```
xernobyl, line 180: **[P2] The harness skips the production middleware chain, so the new RFC 9457 assertions validate a body production never emits.**

(Anchored here; the actual line is `server := httptest.NewServer(mux)` four lines down, which is outside the diff hunk.)

This serves `mux` bare. Production wraps the same mux in `middleware.RequestID(...middleware.ProblemErrors(mux))` — `cmd/api/app.go:682-684`. I ran both chains side by side against a `RespondNotFound` handler:

| | this harness | prod chain |
|---|---|---|
| handler 404 `request_id` | `""`, no `X-Request-ID` header | `019fb759-…` in body **and** header |
| unknown path | `text/plain` · `404 page not found` | `application/problem+json` · `not_found` |
| wrong method | `text/plain` · `Method Not Allowed` | `application/problem+json` · `method_not_allowed` |

Two consequences:

1. The suite structurally **cannot** cover the 404-unknown-path / 405 normalization that `ProblemErrors` exists to provide — those responses never reach it. And the comment at line 121 ("mirrors `cmd/api/app.go`'s taxonomy wiring") is now inaccurate in precisely the dimension this PR is about.
2. **CodeRabbit's suggestion on `tests/taxonomy_api_test.go:79` will fail as written.** `writeProblem` fills `RequestID` from `observability.RequestIDFromContext`, which returns `""` with no `RequestID` middleware in the chain — so `request_id` serializes as `""` (no `omitempty`) and `assert.NotEmpty` fails.

This is pre-existing — the bare `httptest.NewServer(mux)` predates the PR — but it's what the new contract layer now rests on, so it's in scope here.

Suggested order: wrap the harness first —

```go
server := httptest.NewServer(middleware.RequestID(middleware.ProblemErrors(mux)))
```

— then CodeRabbit's `RequestID` assertion holds, and 404-unknown-path / 405 become testable cases worth adding while you're in here.

internal/api/routes/taxonomy.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// RegisterInternalTaxonomy registers the internal taxonomy-service endpoints.
+func RegisterInternalTaxonomy(mux *http.ServeMux, handler *handlers.TaxonomyInternalHandler) {
+	for _, route := range internalTaxonomyRoutes {
+		mux.HandleFunc(route.method+" "+route.path, internalTaxonomyHandler(handler, route.handlerID))
+	}
+}
+
+func publicTaxonomyHandler(handler *handlers.TaxonomyHandler, handlerID publicTaxonomyHandlerID) http.HandlerFunc {
```
xernobyl: **[P2] The handlerID enum + switch indirection reintroduces a silent mis-wire risk the old code didn't have.**

The two `handlerID` enums (lines 10-46) plus these two switch functions spend ~75 lines splitting path and handler into separate lists joined by an enum. The failure mode that creates:

```go
case publicGetTree:
    return handler.GetActiveTree  // compiles; wrong handler
```

That passes `make test-unit`. `taxonomy_test.go` asserts path, `operationId`, and response statuses against the spec — it never checks handler identity — so only the DB-backed integration tests would catch it. The code this replaces (`protected.HandleFunc("GET .../tree", taxonomy.GetTree)`) kept the pair on one line, where a swap is visible at a glance.

I checked all 15 mappings individually and every one is correct today, so this is a design point, not a defect.

Method expressions collapse the whole thing — I compiled and vetted this shape against the real handler types:

```go
type publicTaxonomyRoute struct {
    method           string
    path             string
    operationID      string
    responseStatuses []string
    handler          func(*handlers.TaxonomyHandler, http.ResponseWriter, *http.Request)
}

{
    method: http.MethodGet, path: "/v1/taxonomy/fields",
    operationID: "list-taxonomy-fields",
    responseStatuses: []string{"200", "400", "401", "503", "default"},
    handler: (*handlers.TaxonomyHandler).ListFields,
},
```

```go
func RegisterPublicTaxonomy(mux *http.ServeMux, h *handlers.TaxonomyHandler) {
    for _, route := range publicTaxonomyRoutes {
        mux.HandleFunc(route.method+" "+route.path, func(w http.ResponseWriter, r *http.Request) {
            route.handler(h, w, r)
        })
    }
}
```

That deletes both enums, both switch functions, and both `panic()` branches — which are also unreachable-but-uncovered lines today.

internal/api/routes/taxonomy_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+type openAPIDocument struct {
+	Paths map[string]map[string]openAPIOperation `yaml:"paths"`
```
xernobyl: **[P3] This decode shape breaks on any non-operation path-item key, anywhere in the spec.**

`map[string]map[string]openAPIOperation` assumes every key under a path item is an HTTP method. I checked all 30 path items in `openapi.yaml` — that holds today, so the test passes. But a path-level `summary:`/`description:` (string) or `parameters:` (sequence) on *any* path, taxonomy or not, makes `yaml.Unmarshal` fail for the whole document. Both tests in this file then die on the `require.NoError` in `loadOpenAPI` with a message about an unrelated path — the confusing-failure mode this suite is meant to prevent.

The file already has `isHTTPMethod`. Decoding into `map[string]yaml.Node` and only `Decode`-ing into `openAPIOperation` for keys that pass that check removes the trap without adding much.

tests/taxonomy_api_test.go
```diff
@@ -115,6 +176,21 @@ func TestTaxonomyAPI_PublicReadAndEdit(t *testing.T) {
...
+		requestTaxonomyJSON(ctx, t, http.MethodGet,
+			taxonomyURL(harness.server.URL, "/v1/taxonomy/fields", tenantQuery),
+			harness.apiKey, nil, http.StatusOK, &resp)
+
+		require.Len(t, resp.Data, 1)
+		assert.Equal(t, scope.TenantID, resp.Data[0].TenantID)
+		assert.Equal(t, scope.SourceType, resp.Data[0].SourceType)
+		assert.Equal(t, scope.SourceID, resp.Data[0].SourceID)
+		assert.Equal(t, scope.FieldID, resp.Data[0].FieldID)
+		assert.Equal(t, 2, resp.Data[0].RecordCount)
```
xernobyl: **[P3] Hard-coded count couples this assertion to another helper's internals.**

`2` is `seedTaxonomyGraph`'s 1 record plus `seedEmbeddedFeedback(..., 1)`'s 1. If `seedTaxonomyGraph` ever seeds a second record, this fails here with a misleading message rather than at the cause. Deriving it (or asserting against the seeded IDs' length) keeps the coupling explicit.

Same class as CodeRabbit's `"found 1"` note at line 482 — worth fixing both in one pass.

xernobyl (commented):
The premise checks out — the drift was real: the old harness registered only 4 of the 5 internal taxonomy routes (`heartbeat` was missing entirely, so it had zero coverage), and the shared registry closes that. Both new layers also actually run in CI (`test-unit` covers `./internal/...`, integration runs on pg16/17/18).

Four things worth fixing, in severity order. Two are about the abstraction this PR introduces; one is pre-existing but becomes load-bearing here, because the PR builds a whole RFC 9457 assertion layer on top of it.

I haven't re-raised the three open CodeRabbit threads — all three are valid — but note the interaction flagged in the `taxonomy_support_test.go` comment below: its suggestion #1 (`assert.NotEmpty(t, problem.RequestID)`) turns CI red if applied before the middleware fix.

tests/taxonomy_support_test.go
BhagyaAmarasinghe: Fixed

internal/api/routes/taxonomy_test.go
BhagyaAmarasinghe: Fixed

tests/taxonomy_api_test.go
BhagyaAmarasinghe: Fixed

internal/api/routes/taxonomy.go
BhagyaAmarasinghe: Fixed

xernobyl (approved):
Approving — verified at `aa5cc83`, not just on the resolved threads.

All four of my findings and CodeRabbit's three are genuinely fixed. The method-expression route table drops both enums, both switches and both `panic` branches, and puts the path next to its handler again; I dumped the actual bindings by reflection and all 15 resolve correctly and are mutually distinct. The harness now wraps `middleware.RequestID(middleware.ProblemErrors(mux))`, which is what makes the `request_id` assertion meaningful and the new 404/405 cases reachable at all. The OpenAPI decode tolerates path-item metadata, with a fixture test pinning it.

Ran locally: `make build`, `go vet ./...`, `make test-unit` (0 failures), the full `./tests/...` suite against a migrated pgvector database, `make lint-new` against current main, and `make lint-openapi` — all clean, matching CI.

Worth noting that the contract test paid for itself immediately: ENG-1887 landed on main and added `404` to `list-taxonomy-node-records`, so the table had to be synced. That's the drift mechanism catching a real collision on its first outing rather than a hypothetical one. Nice.

Three non-blocking things, all optional:

1. The new `node records 404 for another tenant` case uses `requestTaxonomyJSON` (status + Content-Type only), while its comment says "Same contract as the run, tree, rename and remove cases above" — and those four all use `requestTaxonomyProblem`, which also checks `code`, `type`, `instance` and `request_id`. I confirmed the stronger helper passes here, so it's a two-line upgrade if you want the comment to be literally true:

```go
requestTaxonomyProblem(ctx, t, http.MethodGet, recordsURL, harness.apiKey, nil,
	http.StatusNotFound, response.CodeNotFound, response.ProblemTypeNotFound)
```

2. `writeProblem` mirrors `request_id` into the `X-Request-ID` header specifically so the two stay consistent "even if upstream middleware ordering changes" — asserting `problem.RequestID == resp.Header.Get("X-Request-ID")` in the shared helper would lock that down now that the middleware is in the harness.

3. `TestTaxonomyAPI_RoutingProblems` is the only test function in the file without a doc comment.

One thing to sort before merge: the description still has `## Summary` / `## Validation` rather than the repo template's `## What does this PR do?` / `## How should this be tested?` plus the checklist. The Validation section does cover the commands, so it's mostly a copy-paste of the reproduction steps into the right sections and ticking the boxes.

PR #119, opened 2026-08-04
feat: add taxonomy lifecycle observability

BhagyaAmarasinghe:
## Summary

- add bounded taxonomy lifecycle, duration, dispatch-failure, and reaper metrics
- emit correlated structured lifecycle logs with optional JSON formatting
- accept optional typed failure diagnostics and persist them in the existing run metrics JSON
- keep the public API and database schema unchanged

## Verification

- `golangci-lint run`
- `go test ./internal/observability ./internal/service ./internal/api/handlers ./internal/repository -count=1`
- taxonomy persistence integration tests against a temporary Postgres instance

## Rollout

This is PR 1 of 5 for [ENG-1200](https://linear.app/formbricks/issue/ENG-1200/add-observability-for-taxonomy-run-duration-failures-and-record-counts). Merge and deploy this compatibility contract before the Taxonomy and GitOps changes.

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -972,7 +972,8 @@ func runTaxonomyRunReaper(
 					started = *run.StartedAt
 				}
 
-				metrics.RecordRunDuration(ctx, time.Since(started), string(models.TaxonomyRunStatusFailed),
+				duration := max(run.FinishedAt.Sub(started), 0)
+				metrics.RecordRunDuration(ctx, duration, string(models.TaxonomyRunStatusFailed),
 					string(run.ScopeType))
 			}
 		}
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -236,10 +236,6 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
 		}
 	}
 
-	// Install TraceContextHandler unconditionally so request_id (and trace_id/span_id when tracing is on) appear in logs.
-	defaultHandler := slog.Default().Handler()
-	slog.SetDefault(slog.New(observability.NewTraceContextHandler(defaultHandler)))
-
 	if tracerProvider != nil {
 		otel.SetTracerProvider(tracerProvider)
 	}
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -550,11 +550,17 @@ func NewApp(cfg *config.Config, db *pgxpool.Pool) (*App, error) {
 		taxonomyStarter = taxonomyClient
 	}
 
+	var taxonomyMetrics observability.TaxonomyMetrics
+	if metrics != nil {
+		taxonomyMetrics = metrics.Taxonomy
+	}
+
 	taxonomyService := service.NewTaxonomyService(service.NewTaxonomyServiceParams{
 		Repo:                  taxonomyRepo,
 		Starter:               taxonomyStarter,
 		EmbeddingModel:        taxonomyEmbeddingModel,
 		MinimumEmbeddingCount: cfg.Taxonomy.MinimumEmbeddedRecords,
+		Metrics:               taxonomyMetrics,
 	})
 	taxonomyHandler := handlers.NewTaxonomyHandler(taxonomyService)
 	feedbackRecordsHandler := handlers.NewFeedbackRecordsHandler(feedbackRecordsService)
@@ -721,7 +727,7 @@ func (a *App) Run(ctx context.Context) error {
 	// Reap taxonomy runs orphaned in a non-terminal state, but only when the taxonomy service is wired
 	// (no runs exist otherwise, so the sweep would be pointless).
 	if a.taxonomyRepo != nil && (a.cfg.Taxonomy.ServiceURL != "" || a.cfg.Taxonomy.ServiceToken != "") {
-		go runTaxonomyRunReaper(ctx, a.taxonomyRepo,
+		go runTaxonomyRunReaper(ctx, a.taxonomyRepo, taxonomyMetricsFromAggregate(a.metrics),
 			a.cfg.Taxonomy.StuckRunTimeout.Duration(), a.cfg.Taxonomy.ReaperInterval.Duration())
 	}
 
@@ -928,7 +934,8 @@ const stuckTaxonomyRunMessage = "taxonomy run timed out without completing"
 // forever in the UI and generation can be retried. Idempotent: the repository's status filter skips
 // runs that finished on their own between sweeps.
 func runTaxonomyRunReaper(
-	ctx context.Context, repo *repository.TaxonomyRepository, timeout, interval time.Duration,
+	ctx context.Context, repo *repository.TaxonomyRepository, metrics observability.TaxonomyMetrics,
+	timeout, interval time.Duration,
 ) {
 	// A non-positive interval panics time.NewTicker, and a non-positive timeout would reap active
 	// runs (cutoff would be now or in the future). Either is misconfiguration — disable the reaper.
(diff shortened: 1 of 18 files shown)


cmd/api/app.go
```diff
@@ -943,10 +950,31 @@ func runTaxonomyRunReaper(
...
+					string(models.TaxonomyRunFailureCodeInternalError), string(run.ScopeType))
+
+				started := run.CreatedAt
+				if run.StartedAt != nil {
+					started = *run.StartedAt
+				}
+
+				metrics.RecordRunDuration(ctx, time.Since(started), string(models.TaxonomyRunStatusFailed),
+					string(run.ScopeType))
+			}
```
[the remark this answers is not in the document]
BhagyaAmarasinghe: Fixed in 15f43df. FailRunIfStale now returns the database finished_at value, ReapedTaxonomyRun carries it, and the reaper computes a non-negative duration from that per-run terminal timestamp.

BhagyaAmarasinghe:
Addressed the outside-diff CodeRabbit request in 15f43df: the Postgres integration suite now verifies that nil failure metrics persist and round-trip as `{}`, and that a known non-empty diagnostics JSON value round-trips through both `MarkRunFailed` and `GetRunForInternalService`. The focused taxonomy API and persistence integration tests pass against pgvector/Postgres 18.

internal/observability/logging.go
```diff
@@ -71,3 +73,27 @@ func (h *TraceContextHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
...
+		logLevel = slog.LevelWarn
+	case "error":
+		logLevel = slog.LevelError
+	default:
+		logLevel = slog.LevelInfo
+	}
+
+	options := &slog.HandlerOptions{Level: logLevel}
+	if strings.EqualFold(strings.TrimSpace(format), "json") {
+		return NewTraceContextHandler(slog.NewJSONHandler(output, options))
```
xernobyl, line 107: **Duplicate `request_id` / `trace_id` / `span_id` on every hub-api log record.**

`NewLogHandler` wraps its output in a `TraceContextHandler` here (and again in the text branch below). But `cmd/api/app.go:239-241` still does:

```go
defaultHandler := slog.Default().Handler()
slog.SetDefault(slog.New(observability.NewTraceContextHandler(defaultHandler)))
```

By that point `slog.Default().Handler()` *is* already a `TraceContextHandler`, so it gets wrapped a second time — and `Handle` appends rather than replaces (`r.AddAttrs`, logging.go:50-58), so both layers add the same keys.

`NewApp` is only ever called from `cmd/api/main.go:58`, which runs after `setupLogging` at `main.go:96`, so this is unconditional for hub-api rather than an edge case. Reproduced by wiring the two call sites exactly the way `cmd/api` does:

```json
{"time":"...","level":"INFO","msg":"taxonomy run started","event":"hub.taxonomy.run.started","request_id":"req-abc","request_id":"req-abc"}
```

`trace_id` and `span_id` double up the same way once tracing is enabled. hub-worker is unaffected — it doesn't re-wrap.

Since `NewLogHandler` now guarantees the wrap for both binaries, the fix is to drop `cmd/api/app.go:239-241` — that block's stated intent ("install TraceContextHandler unconditionally") is satisfied by this function now.

Worth settling before PRs 3-5 build dashboards and queries on these fields.

internal/models/taxonomy.go
```diff
@@ -251,8 +251,26 @@ type TaxonomyRunResultRequest struct {
...
+	ProviderAdapter    string             `json:"provider_adapter,omitempty"        validate:"omitempty,no_null_bytes,max=100"`
+	AdapterVersion     string             `json:"adapter_version,omitempty"         validate:"omitempty,no_null_bytes,max=100"`
+	ProviderSDKVersion string             `json:"provider_sdk_version,omitempty"    validate:"omitempty,no_null_bytes,max=100"`
+	LLMAttempts        *int               `json:"llm_attempts,omitempty"            validate:"omitempty,min=0,max=1000"`
+	InputTokens        *int64             `json:"input_tokens,omitempty"            validate:"omitempty,min=0"`
+	OutputTokens       *int64             `json:"output_tokens,omitempty"           validate:"omitempty,min=0"`
+	TotalTokens        *int64             `json:"total_tokens,omitempty"            validate:"omitempty,min=0"`
+	PhaseDurations     map[string]float64 `json:"phase_durations_seconds,omitempty" validate:"omitempty,max=8,dive,keys,oneof=input_fetch input_validation clustering evidence_selection cluster_labeling taxonomy_generation payload_validation persistence,endkeys,gte=0"` //nolint:lll
```
xernobyl, line 273: **The two phase vocabularies in this struct don't overlap.**

`Phase` (line 262) accepts:

```
fetch cluster label tree persist unknown
```

`PhaseDurations` keys accept:

```
input_fetch input_validation clustering evidence_selection cluster_labeling taxonomy_generation payload_validation persistence
```

Same concept, disjoint sets — no value is legal in both. A producer that reports `phase: "cluster"` cannot key its durations map with `"cluster"`, and one that reports `phase_durations_seconds: {"clustering": 1.2}` cannot use `"clustering"` as its `phase`. Whichever way the taxonomy service picks, one of the two fields gets rejected by the validator or silently reshaped to fit.

Not introduced by the bounding fix — the wider key set is the right call, and I'd keep it as the canonical vocabulary and widen `Phase` to match rather than the reverse. Worth settling here since PR 2/5 is the taxonomy-service side that emits both fields, and this struct is the contract it codes against.

internal/observability/names.go
```diff
@@ -54,6 +54,12 @@ const (
 
 	MetricNameCacheHits   = "hub_cache_hits_total"
 	MetricNameCacheMisses = "hub_cache_misses_total"
+
+	MetricNameTaxonomyRunsStarted   = "hub_taxonomy_runs_started_total"
+	MetricNameTaxonomyRunOutcomes   = "hub_taxonomy_runs_total"
```
xernobyl: **`hub_taxonomy_runs_total` breaks this file's own naming convention.**

Every other terminal-outcome counter here is `hub_<domain>_outcomes_total`:

- `MetricNameEmbeddingOutcomes` → `hub_embedding_outcomes_total` (line 30)
- `MetricNameTranslationOutcomes` → `hub_translation_outcomes_total` (line 37)
- `MetricNameSentimentOutcomes` → `hub_sentiment_outcomes_total` (line 44)
- `MetricNameEmotionsOutcomes` → `hub_emotions_outcomes_total` (line 51)

This one is `MetricNameTaxonomyRunOutcomes` → `hub_taxonomy_runs_total`. The Go constant follows the convention; the emitted metric name doesn't.

It also reads wrong next to its neighbour: `hub_taxonomy_runs_total` sounds like "all taxonomy runs" while it only counts runs that reached a terminal state, and it sits directly above `hub_taxonomy_runs_started_total`. Anyone writing a query will reasonably expect `runs_total >= runs_started_total`, which isn't what these mean.

`hub_taxonomy_outcomes_total` would match the other four. Free to change now; a rename is much more annoying once PRs 3-5 land the GitOps dashboards and alerts against this name.

internal/observability/logging.go
BhagyaAmarasinghe: fixed

internal/models/taxonomy.go
BhagyaAmarasinghe: fixed

internal/observability/names.go
BhagyaAmarasinghe: fixed

xernobyl (approved):
All three points are addressed, and the fixes look right to me:

- the second `TraceContextHandler` wrap is gone from `NewApp`, so `SetupLogging` is the single wrap point for both binaries now — and the count assertion added to `logging_test.go` keeps it from creeping back
- `phase` and `phase_durations_seconds` share one vocabulary
- `hub_taxonomy_outcomes_total` lines up with the other `_outcomes_total` counters

Checked locally at affdd306: build, vet and `go test ./internal/...` all green, and `request_id` now appears exactly once in the JSON output. The per-run `finished_at` for the reaper duration is a nicer fix than what I'd have suggested.

Approving 👍

PR #120, opened 2026-08-07
feat: add filters and sort control to the feedback-record list (ENG-2059)

xernobyl:
## What does this PR do?

Makes the feedback-record list usable at scale. `GET /v1/feedback-records` and `/count` gain 15 new
filters plus sort control. Both endpoints share one filter struct, so every filter applies
identically to both and a count always describes the list it is counting.

Hub half of [ENG-2059](https://linear.app/formbricks/issue/ENG-2059). The Formbricks gateway and MCP
wiring is a follow-up PR — it cannot land until this ships in a Hub release and the
`@formbricks/hub` SDK regenerates from the spec. [ENG-2060](https://linear.app/formbricks/issue/ENG-2060)
builds the filter UI on top of this.

### Filter semantics: AND across filters, OR within a filter

This is the part to be explicit about, because it is the whole ergonomic contract:

| Request | Result | |
|---|---|---|
| `?source_type=survey&source_type=review` | both | repeated values **OR** |
| `?source_type=survey&has_sentiment=true` | intersection | different filters **AND** |

So `?source_type=survey&source_type=review&sentiment=negative` reads as
*"(survey OR review) AND negative"* — a conjunction of disjunctions, which is exactly the shape a
filter bar produces. Each repeatable filter becomes one `col = ANY($n)`; `emotions` uses array
overlap `&&`, which is also ANY.

**What this deliberately cannot express** is OR *across* different fields — "negative sentiment OR
rating ≤ 2". That belongs to saved segments rather than a filter bar. If it is ever needed, the
shape is `POST /v1/feedback-records/query` taking a JSON tree, compiled through the same core so
list/count/query cannot drift. Not built here.

### Filters

| | |
|---|---|
| Repeatable now | `submission_id` `source_type` `source_id` `field_id` `field_group_id` `field_type` `value_id` `user_id` |
| New | `source_name` `language` `created_since`/`created_until` `value_number_min`/`_max` `value_date_min`/`_max` `sentiment` `emotions` `sentiment_score_min`/`_max` `has_sentiment` `has_emotions` `has_translation` |
| Sort | `sort=collected_at\|created_at`, `order=asc\|desc` |

A few decisions worth knowing:

- **A single occurrence behaves exactly as before.** `col = ANY($1)` over a one-element array is the
  same `ScalarArrayOpExpr` the planner produced for `col = $1`; a test pins the legacy placeholder
  layout so this stays true.
- **`created_at` is not `collected_at`.** They diverge on a historical re-import — old
  `collected_at`, today's `created_at` — so "what did this import bring in" is a `created_at`
  question and needs its own pair.
- **`emotions` matches ANY, not ALL.** Two selected chips read as "anger or fear".
- **`field_type` binds through `::text[]::field_type_enum[]`.** The column is a PostgreSQL ENUM and
  pgx has no codec for `field_type_enum[]`; without the cast it works only via a reflection fallback
  that resolves `[]string` to `_text` by coincidence.
- **Every range filter excludes NULL rows** by SQL semantics. `value_number_min=0` selects only
  records that carry a number at all; `sentiment_score_min=-1` means "all *enriched* records". Each
  parameter description says so.
- **`updated_at` is deliberately not a sort key.** The enrichment workers and every PATCH bump it, so
  a row can move across the cursor between pages and be silently dropped. A change feed needs an
  append-only sequence, not a sort parameter.

### ⚠️ Behavior changes

1. **An inverted range is now a 400 instead of an empty 200.** This includes the pre-existing
   `since`/`until` pair, which previously returned an empty page when supplied backwards. An
   inverted range can only ever match zero rows, so it is a client bug, and a 400 naming the pair
   beats a page the caller reads as "there is no such feedback".
2. **`buildFilterConditions` now fails closed on a missing tenant.** It previously emitted *no tenant
   predicate at all* when `tenant_id` was nil — a query spanning every tenant, prevented only by the
   HTTP layer's `required` tag, which workers and tests bypass. Reachable in practice: a
   whitespace-only `tenant_id` satisfies `required` and is caught by this guard.
3. **`submission_id`, `source_type`, `source_id`, `field_id`, `field_group_id` and `user_id` filter
   values are now capped at 255 characters** (previously unbounded). A longer value now returns 400
   instead of matching nothing. `source_type` and `source_name` are unbounded `VARCHAR` in the
   schema, so for those two this is an API-level cap rather than a mirror of the column width.

### Cursors

The cursor now records the ordering it was issued under. Presenting it with a different `sort` or
`order` returns 400 rather than a page that looks like a continuation and is not (arbitrary rows
skipped, others repeated, no way for the client to notice).

Cursors already in client hands keep working on the default listing: the new `s`/`o` fields are
`omitempty` and declared after `t`/`i`, so a cursor with no recorded ordering is **byte-identical**
to the old format. A golden test pins that wire format — the package is shared with the webhooks
endpoints, which are untouched.

Such a cursor is **not** treated as a wildcard. It was issued before sort control, so it holds a
position in the only ordering that existed then (`collected_at desc`); it is resolved to exactly
that and refused anywhere else. Accepting it under `sort=created_at` would bind a collected
timestamp to the `created_at` keyset predicate and skip or repeat rows — the same defect the guard
exists to prevent. (Caught in review; see 9d2e77d.)

### Indexes (migration 021)

Three, all `CONCURRENTLY`:

| Index | Why |
|---|---|
| `(tenant_id, created_at DESC, id)` | the `created_at` filter had zero coverage; the shape mirrors 006 so it also serves `ORDER BY created_at DESC, id ASC` with no sort step |
| `(tenant_id, value_date) WHERE NOT NULL` | sparse — only `date` fields populate it |
| `(tenant_id, source_name) WHERE NOT NULL` | `source_id` has had an index since 001; `source_name` never did |

Verified with `EXPLAIN` on 60k rows: `created_at DESC` is a pure index-only scan, and `created_at
ASC` is an index-only scan **backward** plus an incremental sort (presorted key `created_at`) — so
the mirrored ASC index is deliberately not created. `feedback_records` already carries ~22 indexes;
the migration documents what is left unindexed (`language`, `value_number`, `sentiment_score`, the
FALSE form of the presence filters) and why. `sentiment_score` in particular: the
`feedback_records_sentiment_pairing` CHECK from 014 makes it and `sentiment` NULL together, so the
existing partial `(tenant_id, sentiment)` index already narrows to the candidate rows whenever the
score filter is combined with a sentiment filter — only the uncombined case is exposed, and that's
not worth a permanent write-path cost on a guess.

### Not in this PR

Free-text search (needs `pg_trgm`, a superuser-gated extension — semantic search covers part of the
need today, though it takes no filters), a facets endpoint for ENG-2060's dropdowns, bulk delete
(ENG-2129), and metadata/JSONB filtering.

### Resource bound

The `max=` caps are the only bound on a repeated query parameter — go-playground/form imposes none
of its own. The remaining exposure is the URL itself, bounded by net/http's `MaxHeaderBytes` (1 MB
default, not overridden here), since values are materialized while decoding and rejected afterwards.

## How should this be tested?

Config: `DATABASE_URL` pointing at a `test_db` with migrations applied, `API_KEY` from `.env`.

**Automated**

```bash
make migrate-validate && make init-db   # applies 021
make fmt && make lint                   # 0 issues
make test-unit
make tests                              # integration, needs the DB
make lint-openapi                       # spectral
# with `make run` in another terminal:
make schemathesis
```

New-code coverage is **98.1%** (252/257 added statements); every file but `validation.go` is at 100%,
and its remainder is unreachable defensive branches matching the pre-existing validators beside them.

Integration tests worth looking at (`tests/feedback_record_filters_test.go`):

- `field_type` multi-value against a real DB — settles the enum cast, which no string assertion can
- `emotions` returns a `{joy}` record *and* a `{joy,anger}` record for `?emotions=joy&emotions=anger`
- every range filter includes a row sitting exactly on the bound
- a record with old `collected_at` + new `created_at` is matched by `created_since` and not `since`
- every new filter run against two tenants holding identical data
- all four `(sort, order)` combinations traverse 25 records with deliberate `collected_at` ties in
  pages of four, asserting the result is a permutation of the seeded set — nothing skipped, nothing
  repeated, globally ordered across page boundaries. Without the ties, a broken tiebreak passes.

**Manual smoke test — 52/52**

Four records in a throwaway tenant (`q1` text/enriched, `q2` rating/enriched, `q3` nps, `q4` date):

```bash
# OR within a filter
curl -sG "$HUB/v1/feedback-records" -H "Authorization: Bearer $API_KEY" \
  -d tenant_id=$T -d source_type=survey -d source_type=review     # -> q1,q2,q3

# AND across filters
curl -sG "$HUB/v1/feedback-records" -H "Authorization: Bearer $API_KEY" \
  -d tenant_id=$T -d source_type=survey -d has_sentiment=true     # -> q1

# emotions is ANY, not ALL
curl -sG "$HUB/v1/feedback-records" -H "Authorization: Bearer $API_KEY" \
  -d tenant_id=$T -d emotions=joy -d emotions=anger               # -> q1,q2
```

| Group | Result |
|---|---|
| 12 identity/multi-value filters | ✅ |
| 4 date filters (`since`/`until`/`created_*`) | ✅ |
| 4 numeric & date ranges, inclusive on both bounds | ✅ |
| 10 enrichment filters (sentiment, emotions, score, `has_*`) | ✅ |
| 4 sort combinations + default unchanged | ✅ |
| 3 `/count` parity checks | ✅ |
| 9 rejections → 400 (bad enum, bad sort, inverted range, out-of-range score) | ✅ |
| cursor page 2, and cursor carried to another ordering → 400 | ✅ |
| tenant isolation, missing/blank `tenant_id` | ✅ |

Representative 400 bodies:

```
sentiment           | must be one of: very_negative, negative, neutral, positive, very_positive, mixed
value_number_min    | must be less than or equal to value_number_max
cursor              | was issued for a different sort/order; restart pagination without a cursor,
                      or keep sort and order unchanged
tenant_id           | tenant_id is required to list feedback records   (blank tenant, ?tenant_id=%20%20)
```

**Review round**

Five review comments addressed in 9d2e77d, one of them a real defect (the legacy-cursor hole above).
Smoke suite re-run afterwards: still 52/52. The behaviours the fixes changed were then checked over
HTTP, since the suite does not cover them:

| Request | Status | |
|---|---|---|
| legacy cursor, default ordering | 200 | old cursors keep working |
| legacy cursor, `sort=created_at` | **400** | the fix — previously 200, with wrong rows |
| legacy cursor, `order=asc` | **400** | the fix, on the order axis too |
| legacy cursor, explicit `sort=collected_at&order=desc` | 200 | naming the default == omitting it |
| sort-tagged cursor, same ordering | 200 | unchanged |
| `submission_id` 255 / 256 chars | 200 / **400** | new per-element cap, boundary included |

**Security**

A security review found one real defect, fixed here: go-playground/form consults its custom type
funcs only for the plain query key, so the *indexed* form (`?field_type[0]=bogus`) skipped them and
assigned raw strings. Values were always bound as parameters — never SQL injection, never a tenant
crossing — but `field_type` then failed the enum cast at Postgres as an unmapped 500, and
`sentiment`/`emotions` were accepted silently. Fixed with `dive` element validators, which run on
the decoded struct however it was populated; regression-tested.

Also verified: all identifiers in generated SQL come from typed constants (`sqlColumn`/`sqlOperator`)
with an exhaustive fail-closed allowlist for `sort`; `gitleaks` and `semgrep` (`p/golang`,
`p/sql-injection`) clean on every changed file; `govulncheck` 0.

### Required

- [x] Ran `make build`

### Appreciated

- [ ] Updated docs in `docs/` if changes were necessary — no `docs/` changes needed; the OpenAPI spec is the reference

---

**Deploy order: migration first, then code.** Serving `sort=created_at` before
`CREATE INDEX CONCURRENTLY` finishes is exactly the residual scan this removes.


internal/service/feedback_records_service.go
```diff
@@ -327,12 +337,19 @@ func (s *FeedbackRecordsService) ListFeedbackRecords(
...
 
-		records, hasMore, err = s.repo.ListAfterCursor(ctx, filters, collectedAt, id)
+		// A cursor bounds one column in one direction. Under a different ordering it still yields
+		// a valid-looking page that is not the continuation of the previous one, so refuse it
+		// rather than silently skipping and repeating rows.
+		if matchErr := key.Match(string(filters.Sort), string(filters.Order)); matchErr != nil {
+			return nil, fmt.Errorf("validate cursor ordering: %w", matchErr)
+		}
+
+		records, hasMore, err = s.repo.ListAfterCursor(ctx, filters, key.Timestamp, key.ID)
```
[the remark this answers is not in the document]
xernobyl: Good catch, and this was the important one — fixed in 9d2e77d.

You are right about the mechanism: a metadata-free cursor is not orderless, it holds a `collected_at DESC` position, so accepting it under `sort=created_at` binds that timestamp to the wrong keyset column. That is exactly the failure the cursor rework exists to prevent, and I left the hole in it.

Rather than special-casing inside `Match`, I made `Match` compare exactly and moved the legacy rule to the caller as `Key.ResolveOrdering(defaultSort, defaultOrder)`. `pkg/cursor` is shared with the webhooks endpoints, so keeping endpoint policy out of it matters; the service is the layer that knows what this endpoint's original ordering was.

Old cursors keep working on the default listing and are refused anywhere else. Regression tests at all three levels: `TestKey_ResolveOrdering`, `TestListFeedbackRecords_RejectsCursorFromAnotherOrdering/a_legacy_cursor_is_refused_under_a_non-default_ordering`, and the same case in `TestFeedbackRecordFilters_CursorIsBoundToItsOrdering`. The OpenAPI `cursor` description no longer claims legacy cursors "remain valid" unqualified.

openapi.yaml
```diff
@@ -140,7 +180,41 @@ paths:
...
+                        malformed cursor, or a cursor issued for a different sort/order.
+                    content:
+                        application/problem+json:
+                            schema:
+                                $ref: '#/components/schemas/ErrorModel'
+                            examples:
+                                inverted_range:
+                                    summary: A range filter supplied backwards
+                                    value:
+                                        type: "https://formbricks.com/docs/errors/validation"
```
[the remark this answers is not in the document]
xernobyl: Fixed in 9d2e77d. Both examples now use `https://hub.formbricks.com/problems/validation`.

Worth noting this was not only an internal inconsistency — it was factually wrong. The server emits `hub.formbricks.com/problems/validation`, which I confirmed against a live response while smoke-testing, so the examples I added did not match reality.

internal/models/feedback_records_filters.go
```diff
@@ -0,0 +1,203 @@
...
+//
+// TenantID is deliberately NOT repeatable. It is the tenant isolation boundary; every other filter
+// narrows within it.
+type ListFeedbackRecordsFilters struct {
+	TenantID *string `form:"tenant_id" validate:"required,no_null_bytes,min=1"`
+
+	// Repeatable identity filters. A single occurrence behaves exactly as it did when these were
+	// scalars: the repository emits `col = ANY($n)` over a one-element array, which the planner
+	// treats as the same ScalarArrayOpExpr an equality produces.
+	SubmissionID []string    `form:"submission_id"  validate:"omitempty,max=100,dive,no_null_bytes"`
```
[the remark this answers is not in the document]
xernobyl: Fixed in 9d2e77d, both the tag and the OpenAPI items schema.

The reasoning holds: `submission_id` is stored as `VARCHAR(255)` and `CreateFeedbackRecordRequest` caps it at 255, so a longer filter value can never match a row — it only widens the bound array.

internal/models/feedback_records_filters.go
```diff
@@ -0,0 +1,203 @@
...
+	appendInvertedTimeRange(&inverted, f.Since, f.Until, "Since", "since", "until")
+	appendInvertedTimeRange(&inverted, f.CreatedSince, f.CreatedUntil, "CreatedSince", "created_since", "created_until")
+	appendInvertedTimeRange(&inverted, f.ValueDateMin, f.ValueDateMax, "ValueDateMin", "value_date_min", "value_date_max")
+	appendInvertedFloatRange(
+		&inverted, f.ValueNumberMin, f.ValueNumberMax, "ValueNumberMin", "value_number_min", "value_number_max",
+	)
+	appendInvertedFloatRange(
+		&inverted, f.SentimentScoreMin, f.SentimentScoreMax,
+		"SentimentScoreMin", "sentiment_score_min", "sentiment_score_max",
+	)
```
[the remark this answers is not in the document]
xernobyl: Fixed in 9d2e77d — five pairs, capacity 5, comment corrected.

tests/feedback_record_filters_test.go
```diff
@@ -0,0 +1,590 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			with := env.list(t, func(f *models.ListFeedbackRecordsFilters) { tt.set(f, &present) })
+			without := env.list(t, func(f *models.ListFeedbackRecordsFilters) { tt.set(f, &absent) })
+
+			assert.Equal(t, []uuid.UUID{tt.wantWithID}, with)
+			assert.Equal(t, []uuid.UUID{tt.wantoutID}, without)
+			assert.Len(t, append(with, without...), env.count(t, nil), "the two sides must partition the tenant")
+		})
+	}
+}
```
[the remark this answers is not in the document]
xernobyl: Fixed in 9d2e77d, and thanks — chasing this turned up a weakness in the test beyond the missing case.

Adding `has_translation` to the existing table would not have caught anything: the single enriched record carried sentiment, emotions *and* a translation, so all three filters returned the same rows and a predicate reading the wrong column would still have passed. The fixture now has a third record with sentiment and emotions but **no** translation, so the three filters partition differently.

Verified it discriminates by pointing `has_translation` at `colSentiment` and confirming the test goes red, then reverting. `wantoutID` renamed to `wantWithoutID`.

migrations/021_feedback_record_filter_indexes.sql
```diff
@@ -0,0 +1,73 @@
...
+  ON feedback_records (tenant_id, value_date) WHERE value_date IS NOT NULL;
+
+-- source_name is the only new string-equality filter with no index of any kind: source_id has had
+-- one since 001, source_name never did. PARTIAL for the same sparsity reason — it is optional on
+-- ingestion.
+DROP INDEX CONCURRENTLY IF EXISTS idx_feedback_records_tenant_source_name;
+CREATE INDEX CONCURRENTLY idx_feedback_records_tenant_source_name
+  ON feedback_records (tenant_id, source_name) WHERE source_name IS NOT NULL;
+
+-- Deliberately NOT indexed here — feedback_records already carries ~22 indexes and every addition
```
Dhruwang, line 59: **P2 — `sentiment_score` range filter has no supporting index, and this omission note doesn't mention it.**

`sentiment_score_min/_max` builds a `>=`/`<=` range (`filter_conditions.go` `addFloatRange(colSentimentScore, ...)`), but no migration creates an index on `sentiment_score` (grep of `migrations/` confirms — it only appears in column/CHECK DDL). This is the same residual-scan shape this migration deliberately fixed for `created_at`: on a large tenant the planner can only satisfy it by walking `idx_feedback_records_tenant_collected_at_id` and discarding rows.

This block carefully enumerates every *deliberately* unindexed filter (language, value_number, has_* FALSE form) with a rationale — `sentiment_score` is simply absent from both the index set and this note. Either add a partial `(tenant_id, sentiment_score) WHERE sentiment_score IS NOT NULL` index, or add a bullet here explaining why it's left unindexed (as the others do). Lower-traffic than `created_at`, so the fix is a judgement call — but the silent omission is the gap worth closing.

internal/models/feedback_records_filters.go
```diff
@@ -0,0 +1,203 @@
...
+//
+// go-playground/form imposes no useful bound of its own on a repeated query parameter, so these
+// caps are the only thing between a caller and an arbitrarily long IN-list. The remaining exposure
+// is bounded by net/http's MaxHeaderBytes, since values are materialized while decoding and
+// rejected afterwards.
+//
+// Struct tags cannot reference a constant, so every cap below is written out as a literal.
+// TestFilterValueCapsMatchTheirSets reflects over the tags and enforces that they agree with this
+// constant and with the enum sets, which is what keeps adding a label from silently making a
+// legitimate request a 400. Keep the OpenAPI maxItems for these parameters in step by hand.
```
Dhruwang, line 60: **P2 — the OpenAPI `maxItems` ↔ `MaxFilterValues` coupling is enforced in only one direction.**

This comment says to keep the OpenAPI `maxItems` in step by hand, and `TestFilterValueCapsMatchTheirSets` pins the Go caps to their sources — but nothing ties `openapi.yaml`'s `maxItems` back to `MaxFilterValues`/the enum cardinalities (verified: no test references `openapi.yaml`). A future enum addition can leave the published spec understating what the API actually accepts, and `make lint-openapi`/schemathesis won't catch it because fuzzing stays within the (stale) documented bound.

Suggest extending the existing reflection-based test (or a spec-generation check) to also assert against `openapi.yaml`, so the manual step becomes an enforced invariant.

openapi.yaml
```diff
@@ -2166,84 +2262,319 @@ components:
...
             schema:
-                type: string
-                description: Filter by submission ID (e.g. UUID from webhook idempotency). NULL bytes not allowed.
-                pattern: '^[^\x00]*$'
-                example: "550e8400-e29b-41d4-a716-446655440000"
+                type: array
+                maxItems: 100
+                items:
+                    type: string
+                    maxLength: 255
```
Dhruwang, line 2293: **P3 — previously-unbounded single-value filters gained `maxLength: 255` (and matching `max=255` Go validators).**

This is a genuine narrowing of input that used to be unbounded (`submission_id`, `source_type`, `source_id`, `field_id`, `field_group_id`, `user_id`). It's a reasonable cap, but it's a behaviour change a caller could hit, and it isn't called out in the PR's "Behaviour changes" section alongside the inverted-range change. Worth one line there so integrators aren't surprised by a new 400.

(Related: `TestStringFilterElementCapsMatchTheStoredWidth`'s comment labels `source_type`/`source_name` as `VARCHAR(255)`, but migration 001 declares them as bare unbounded `VARCHAR` — the test's stated "cap below stored width" invariant can't actually be violated for those two columns. No functional bug, just an inaccurate comment.)

openapi.yaml
```diff
@@ -2166,84 +2262,319 @@ components:
...
+            in: query
+            description: |
+                Filter by value_date <= value_date_max (ISO 8601, inclusive). Records whose value_date
+                is NULL are excluded.
+            schema:
+                type: string
+                format: date-time
+                example: "2026-12-31T00:00:00Z"
+        FeedbackRecordsSentiment:
+            name: sentiment
```
Dhruwang, line 2509: **P3 — `sentiment` and `emotions` descriptions omit the "empty value is a no-op" note that `field_type` documents.**

`field_type`'s description states that an empty value is ignored rather than rejected (`?field_type=` is a no-op), and `parseEnumValues` gives `sentiment`/`emotions` the identical decode semantics — but their descriptions don't mention it. Mirror the note here so all three fixed-vocabulary filters document the same behaviour.

Dhruwang (commented):
Automated multi-agent review + live API testing (spun up the branch against Postgres, applied migration 021, exercised the endpoint). Correctness, security, tenant-isolation, cursor/keyset behaviour, and the migration all came back clean and verified against a running instance — cross-ordering cursor reuse returns 400 `cursor_sort_mismatch`, inverted ranges and bogus enums (including the indexed `field_type[0]=` form) all 400, no 500s. Nice work.

Five non-blocking points below — two P2, three P3 (doc). Four are inline; one has no diff anchor so it's here:

**P3 — `/v1/feedback-records/count` 400 response doc** wasn't updated to match the newly-detailed 400 on the list endpoint, even though both share the same validation now. Worth mirroring the list endpoint's `400` documentation on `/count` so the two stay consistent.

migrations/021_feedback_record_filter_indexes.sql
xernobyl: Fixed in ba3548e — documented the omission rather than adding an index. Chased down a real reason rather than a guess: the feedback_records_sentiment_pairing CHECK from 014 makes sentiment and sentiment_score NULL together, so the existing partial (tenant_id, sentiment) index already narrows to exactly the candidate rows whenever the score filter is combined with a sentiment filter. Only the uncombined case is left exposed, and a permanent write-path cost on the ingestion path isn't worth paying for that on a guess. Left the ready-to-run CREATE INDEX in the comment for whoever does the EXPLAIN. Also corrected the header above, which said "three filters" — now says four, with sentiment_score called out as the deliberate exception. Added the omission and the index rationale to the PR description too.

internal/models/feedback_records_filters.go
xernobyl: Fixed in ba3548e — added TestOpenAPICapsMatchTheFilterStruct, extending the existing reflection-based approach as suggested, placed right next to TestFilterValueCapsMatchTheirSets in the same package. It checks both bounds that drift for the same reason: maxItems against MaxFilterValues/enum cardinality, and items.maxLength against the per-element cap. Mutation-tested five ways (understated maxItems, understated maxLength, a removed maxItems, an undocumented string filter, and a differently-sized element cap on language to prove the caps are derived per-filter rather than hardcoded) — each fails with a message naming the exact filter and the mismatch.

openapi.yaml
xernobyl: Added the line to the PR description's Behaviour changes section in ba3548e's push. On the sub-point: checked, and it's slightly worse than an inaccurate comment — source_type and source_name are bare unbounded VARCHAR (confirmed against the live DB via information_schema), not VARCHAR(255) like the others. So for those two the 255 cap is a real (if theoretical) narrowing below the stored width, not a free assertion. Rewrote TestStringFilterElementCapsMatchTheStoredWidth's comment to say so rather than repeat the false claim.

openapi.yaml
xernobyl: Fixed in ba3548e — mirrored field_type's exact wording onto sentiment and emotions. Confirmed all three share the identical parseEnumValues behavior (if raw == "" { continue }) before mirroring the note, so the three descriptions now say the same true thing rather than just looking consistent.

xernobyl (commented):
Addressed in 55413e5 — the /count 400 now mirrors the list endpoint's detailed description and inverted_range example, minus cursor-specific content since count has no pagination. Both endpoints now document the same validation failures consistently.

Dhruwang (approved):

PR #121, opened 2026-08-12
feat(embeddings): batch worker requests and control backfills

BhagyaAmarasinghe:
## What does this PR do?

Fixes the Hub-side throughput bottleneck that sent one TEI request per embedding job and left operators without a safe way to resume a stranded taxonomy backfill.

- adds optional OpenAI-compatible array embedding requests with strict response count, index, dimension, and finite-value validation
- coalesces worker document requests into bounded micro-batches while keeping one River job, freshness check, retry, and database write per record
- keeps query embeddings synchronous and preserves single-request fallback for non-batch providers
- adds batch throughput, outcome, duration, and in-flight metrics
- packages `backfill-embeddings` in the production image and adds tenant, count-only, and max-record controls
- publishes a durable `taxonomy_embedding` missing-record gauge so an empty embedding queue with remaining records is alertable

There is no external HTTP API or database-schema change. The new batch settings default to disabled behavior (`EMBEDDING_BATCH_SIZE=1`).

## How should this be tested?

- `make build` ✅
- `make lint GOLANGCI_LINT=/Users/bhagya/work/bin/golangci-lint` ✅ — 0 issues
- `go test -race ./internal/openai ./internal/service ./internal/observability ./internal/workers ./internal/config ./cmd/api ./cmd/worker ./cmd/backfill-embeddings` ✅
- `DATABASE_URL=postgres://postgres:postgres@localhost:55432/test_db?sslmode=disable API_KEY=test-api-key-for-ci go test ./tests/... -count=1 -timeout=120s` ✅ against an isolated `pgvector/pgvector:pg18` database
- Verify a worker configured with batch size 8 emits one ordered vector per River job and query requests remain single.
- Run `backfill-embeddings --taxonomy --count-only --tenant-id <tenant>`, then a bounded `--max-records` canary, before an unlimited enqueue.

### Required

- [x] Ran `make build`
- [x] Ran integration tests in `tests/`
- [x] Ran formatter and full linter; no warnings
- [x] Branch contains the latest `origin/main`
- [x] No database schema change

### Appreciated

- [x] No external API/OpenAPI change
- [x] Updated environment-variable documentation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,10 +40,11 @@ RUN mkdir -p /tmp/migration-tools && \
 COPY go.mod go.sum ./
 RUN go mod download
 
-# Build the application (hub-api and hub-worker)
+# Build the application and controlled embedding backfill utility.
 COPY . .
 RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o /build/bin/hub-api ./cmd/api && \
-    CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o /build/bin/hub-worker ./cmd/worker
+    CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o /build/bin/hub-worker ./cmd/worker && \
+    CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o /build/bin/backfill-embeddings ./cmd/backfill-embeddings
 
 # =============================================================================
 # Stage 2: Runtime (default: hub-api)
@@ -60,6 +61,7 @@ WORKDIR /app
 # Copy binaries and migration tools from builder
 COPY --from=builder /build/bin/hub-api /app/hub-api
 COPY --from=builder /build/bin/hub-worker /app/hub-worker
+COPY --from=builder /build/bin/backfill-embeddings /app/backfill-embeddings
 COPY --from=builder /go/bin/goose /usr/local/bin/goose
 COPY --from=builder /go/bin/river /usr/local/bin/river
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,10 +74,11 @@ USER app
 
 EXPOSE 8080
 
-# Health check for hub-api. Disable or override when running hub-worker (e.g. docker run ... hub-worker)
-# since workers do not expose HTTP.
+# Health check only hub-api. hub-worker and backfill-embeddings do not expose HTTP, so commands
+# that reuse this image must not inherit a failing API probe.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
-	CMD wget --no-verbose --tries=1 --spider http://localhost:8080/health || exit 1
+	CMD [ "$(readlink /proc/1/exe)" != "/app/hub-api" ] || wget --no-verbose --tries=1 --spider http://localhost:8080/health || exit 1
 
-# Default: run hub-api. Override with command to run hub-worker: docker run ... hub-worker
+# Default: run hub-api. Override the entrypoint for another binary, for example
+# docker run --entrypoint /app/hub-worker ... or --entrypoint /app/backfill-embeddings ...
 ENTRYPOINT ["/app/hub-api"]
--- a/cmd/worker/app.go
+++ b/cmd/worker/app.go
@@ -25,6 +25,7 @@ type WorkerApp struct {
 	cfg            *config.Config
 	db             *pgxpool.Pool
 	river          *river.Client[pgx.Tx]
+	embeddingBatch *service.BatchingEmbeddingClient
 	meterProvider  *sdkmetric.MeterProvider
 	tracerProvider *sdktrace.TracerProvider
 }
@@ -102,7 +103,10 @@ func NewWorkerApp(cfg *config.Config, db *pgxpool.Pool) (*WorkerApp, error) {
 		taxonomyEmbeddingEnqueueModel = ""
 	}
 
-	var translationRecordsService *service.FeedbackRecordsService
+	var (
+		translationRecordsService *service.FeedbackRecordsService
+		embeddingBatch            *service.BatchingEmbeddingClient
+	)
 
 	if providerName != "" {
 		embeddingCfg := service.EmbeddingClientConfig{
@@ -125,6 +129,28 @@ func NewWorkerApp(cfg *config.Config, db *pgxpool.Pool) (*WorkerApp, error) {
 			shutdownObservability(context.Background(), meterProvider, tracerProvider)
 
 			return nil, fmt.Errorf("create embedding client: %w", err)
+		}
+
+		if batchClient, enabled := service.NewBatchingEmbeddingClient(
+			embeddingClient,
+			service.EmbeddingBatchConfig{
+				BatchSize:   cfg.Embedding.BatchSize,
+				MaxWait:     time.Duration(cfg.Embedding.BatchMaxWaitMs) * time.Millisecond,
+				MaxInFlight: cfg.Embedding.BatchMaxInFlight,
+			},
+			embeddingMetrics,
+		); enabled {
+			embeddingClient = batchClient
+			embeddingBatch = batchClient
+
+			slog.Info("embedding document batching enabled",
+				"batch_size", cfg.Embedding.BatchSize,
+				"max_wait_ms", cfg.Embedding.BatchMaxWaitMs,
+				"max_in_flight", cfg.Embedding.BatchMaxInFlight,
+			)
+		} else if cfg.Embedding.BatchSize > 1 {
+			slog.Info("embedding document batching unavailable for provider; using single requests",
+				"provider", providerName)
 		}
 
 		feedbackRecordsRepo := repository.NewFeedbackRecordsRepository(db)
@@ -292,6 +318,7 @@ func NewWorkerApp(cfg *config.Config, db *pgxpool.Pool) (*WorkerApp, error) {
 		cfg:            cfg,
 		db:             db,
 		river:          riverClient,
+		embeddingBatch: embeddingBatch,
 		meterProvider:  meterProvider,
 		tracerProvider: tracerProvider,
 	}, nil
@@ -349,6 +376,15 @@ func (a *WorkerApp) Run(ctx context.Context) error {
 
 	<-a.river.Stopped()
 
+	if a.embeddingBatch != nil {
+		batchCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), a.cfg.Server.ShutdownTimeout.Duration())
+		defer cancel()
+
+		if err := a.embeddingBatch.Shutdown(batchCtx); err != nil {
+			return fmt.Errorf("embedding batch shutdown: %w", err)
+		}
+	}
+
 	return nil
 }
 
@@ -372,6 +408,16 @@ func shutdownObservability(ctx context.Context, meter *sdkmetric.MeterProvider, 
 func (a *WorkerApp) Shutdown(ctx context.Context) (err error) {
 	if stopErr := a.river.Stop(ctx); stopErr != nil {
 		err = fmt.Errorf("river stop: %w", stopErr)
+	}
+
+	if a.embeddingBatch != nil {
+		if batchErr := a.embeddingBatch.Shutdown(ctx); batchErr != nil {
+			if err == nil {
+				err = fmt.Errorf("embedding batch shutdown: %w", batchErr)
+			} else {
+				slog.Error("shutdown embedding batcher", "error", batchErr)
+			}
+		}
 	}
 
 	if a.tracerProvider != nil {
--- a/cmd/worker/app.go
+++ b/cmd/worker/app.go
@@ -377,7 +377,7 @@ func (a *WorkerApp) Run(ctx context.Context) error {
 	<-a.river.Stopped()
 
 	if a.embeddingBatch != nil {
-		batchCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), a.cfg.Server.ShutdownTimeout.Duration())
+		batchCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), embeddingBatchFlushTimeout)
 		defer cancel()
 
 		if err := a.embeddingBatch.Shutdown(batchCtx); err != nil {
@@ -391,6 +391,10 @@ func (a *WorkerApp) Run(ctx context.Context) error {
 // riverStopAndCancelTimeout bounds the escalated (job-cancelling) stop after the graceful stop
 // timed out; kept short so the pod exits within the orchestrator's termination grace period.
 const riverStopAndCancelTimeout = 5 * time.Second
+
+// embeddingBatchFlushTimeout bounds the one final partial-batch flush without extending River's
+// graceful shutdown by another full server shutdown period.
+const embeddingBatchFlushTimeout = 5 * time.Second
 
 func shutdownObservability(ctx context.Context, meter *sdkmetric.MeterProvider, tracer *sdktrace.TracerProvider) {
 	if tracer != nil {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -364,7 +364,7 @@ func TestLoad_EmbeddingBaseURL(t *testing.T) {
 func TestLoad_EmbeddingBatchSettings(t *testing.T) {
 	t.Setenv("API_KEY", "test-api-key")
 	t.Setenv("EMBEDDING_BATCH_SIZE", "8")
-	t.Setenv("EMBEDDING_BATCH_MAX_WAIT_MS", "25")
+	t.Setenv("EMBEDDING_BATCH_MAX_WAIT_MS", "37")
 	t.Setenv("EMBEDDING_BATCH_MAX_IN_FLIGHT", "3")
 
 	cfg, err := Load()
@@ -376,8 +376,8 @@ func TestLoad_EmbeddingBatchSettings(t *testing.T) {
 		t.Errorf("Embedding.BatchSize = %d, want 8", cfg.Embedding.BatchSize)
 	}
 
-	if cfg.Embedding.BatchMaxWaitMs != 25 {
-		t.Errorf("Embedding.BatchMaxWaitMs = %d, want 25", cfg.Embedding.BatchMaxWaitMs)
+	if cfg.Embedding.BatchMaxWaitMs != 37 {
+		t.Errorf("Embedding.BatchMaxWaitMs = %d, want 37", cfg.Embedding.BatchMaxWaitMs)
 	}
 
 	if cfg.Embedding.BatchMaxInFlight != 3 {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -361,6 +361,30 @@ func TestLoad_EmbeddingBaseURL(t *testing.T) {
 	}
 }
 
+func TestLoad_EmbeddingBatchSettings(t *testing.T) {
+	t.Setenv("API_KEY", "test-api-key")
+	t.Setenv("EMBEDDING_BATCH_SIZE", "8")
+	t.Setenv("EMBEDDING_BATCH_MAX_WAIT_MS", "25")
+	t.Setenv("EMBEDDING_BATCH_MAX_IN_FLIGHT", "3")
+
+	cfg, err := Load()
+	if err != nil {
+		t.Fatalf("Load() error = %v", err)
+	}
+
+	if cfg.Embedding.BatchSize != 8 {
+		t.Errorf("Embedding.BatchSize = %d, want 8", cfg.Embedding.BatchSize)
+	}
+
+	if cfg.Embedding.BatchMaxWaitMs != 25 {
+		t.Errorf("Embedding.BatchMaxWaitMs = %d, want 25", cfg.Embedding.BatchMaxWaitMs)
+	}
+
+	if cfg.Embedding.BatchMaxInFlight != 3 {
+		t.Errorf("Embedding.BatchMaxInFlight = %d, want 3", cfg.Embedding.BatchMaxInFlight)
+	}
+}
+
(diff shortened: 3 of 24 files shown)


internal/service/feedback_records_service.go
```diff
@@ -597,6 +600,51 @@ func (s *FeedbackRecordsService) BackfillEmbeddingsWithInputKind(
...
+				ctx, tenantID, model, inputKind, afterID, limit)
+		})
+}
+
+func (s *FeedbackRecordsService) backfillEmbeddingsWithInputKind( //nolint:funcorder // shared helper stays with public backfill entrypoints
+	ctx context.Context,
+	model string,
+	inputKind models.EmbeddingInputKind,
+	maxRecords int,
+	listPage func(context.Context, uuid.UUID, int) ([]uuid.UUID, error),
```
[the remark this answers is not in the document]
xernobyl: Checked this against the code — I don't think this holds up, so I'm not asking for a change here.

`tenant_id` is immutable on `feedback_records` (see the comment at `feedback_records_repository.go:154`: "tenant_id is immutable on feedback records"), so a record can never move between tenants after creation. `ListTenantFeedbackRecordIDsForBackfillByInputKind` filters candidates with a bound `tenant_id` parameter, and once an ID is selected inside tenant A's page it can never later belong to tenant B — there's no window for the tenant to shift under the job. This mirrors how the existing event-triggered embedding path already works: the record ID itself is the authorization boundary, with no tenant field on `FeedbackEmbeddingArgs` at all.

`TestTenantEmbeddingBackfillStaysInsideTenant` already covers this end to end (repository listing + service enqueue, both asserted to never touch the foreign tenant's record).

Flagging as verified-non-issue rather than fixing, so this doesn't get re-raised without this context in a future pass.

internal/service/embedding_batcher.go
```diff
@@ -0,0 +1,314 @@
...
+	status := "success"
+	if err != nil {
+		status = "error"
+	}
+
+	if b.metrics != nil {
+		b.metrics.RecordEmbeddingBatch(context.WithoutCancel(providerCtx), int64(len(inputs)), time.Since(started), status)
+	}
+
+	b.deliver(active, vectors, err)
```
xernobyl, line 265: [P2] Confirmed. `deliver` fans the same `err` out to every request in `active` when `CreateEmbeddings` fails, so one bad/rejected input (blank text, a per-input token-limit rejection, provider content policy) fails every unrelated document sharing that batch window. Each collateral record then burns a River retry attempt, and since batches re-form differently on each retry, a persistently bad record can keep damaging different peers across attempts and exhaust `EMBEDDING_MAX_ATTEMPTS` on records that were never the problem.

Not a blocker for this PR — `EMBEDDING_BATCH_SIZE` defaults to `1` (batching disabled), so this doesn't fire until an operator turns batching on. Worth adding the individual-retry fallback (CodeRabbit's sketch is reasonable) before enabling batching in production.

cmd/worker/app.go
```diff
@@ -349,6 +376,15 @@ func (a *WorkerApp) Run(ctx context.Context) error {
...
 
+	if a.embeddingBatch != nil {
+		batchCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), a.cfg.Server.ShutdownTimeout.Duration())
+		defer cancel()
+
+		if err := a.embeddingBatch.Shutdown(batchCtx); err != nil {
+			return fmt.Errorf("embedding batch shutdown: %w", err)
+		}
+	}
+
```
xernobyl, line 387: [P3] Confirmed. This grants the batch flush a full `ShutdownTimeout` on top of River's own stop budget plus the escalation timeout, so total graceful shutdown can exceed what `terminationGracePeriodSeconds` expects and risk a SIGKILL mid-flush. The final flush only ever handles one partial batch — give it its own small explicit budget (e.g. 5s) instead of reusing `ShutdownTimeout`.

xernobyl (commented):
House-style pass on top of CodeRabbit's review. Verified each candidate against the current head; details inline. The tenant-boundary finding on `feedback_records_service.go` is a verified false positive (immutable `tenant_id` + existing test coverage) — replied in that thread rather than filing a change. No merge blockers.

Dockerfile
```diff
@@ -40,10 +40,11 @@ RUN mkdir -p /tmp/migration-tools && \
 COPY go.mod go.sum ./
 RUN go mod download
 
-# Build the application (hub-api and hub-worker)
+# Build the application and controlled embedding backfill utility.
 COPY . .
 RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o /build/bin/hub-api ./cmd/api && \
-    CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o /build/bin/hub-worker ./cmd/worker
+    CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o /build/bin/hub-worker ./cmd/worker && \
+    CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -o /build/bin/backfill-embeddings ./cmd/backfill-embeddings
```
xernobyl, line 47: [P3] Confirmed, but not a new gap: the inherited `/health` HEALTHCHECK already required override for `hub-worker` (documented in the comment above) since workers don't serve HTTP. `backfill-embeddings` now shares that same caveat and isn't mentioned. Low risk since it's a one-off CLI rather than a long-running deployment, but worth extending the comment so it's not missed operationally.

internal/config/config_test.go
```diff
@@ -361,6 +361,30 @@ func TestLoad_EmbeddingBaseURL(t *testing.T) {
...
+		t.Fatalf("Load() error = %v", err)
+	}
+
+	if cfg.Embedding.BatchSize != 8 {
+		t.Errorf("Embedding.BatchSize = %d, want 8", cfg.Embedding.BatchSize)
+	}
+
+	if cfg.Embedding.BatchMaxWaitMs != 25 {
+		t.Errorf("Embedding.BatchMaxWaitMs = %d, want 25", cfg.Embedding.BatchMaxWaitMs)
+	}
```
xernobyl, line 381: [P4] Confirmed trivial nit. `EMBEDDING_BATCH_MAX_WAIT_MS` is set to `25`, which is also `applyDefaults`'s fallback value, so this assertion would still pass even if the env tag were dropped or misspelled. Pick a non-default value per CodeRabbit's suggested diff.

internal/service/embedding_client.go
```diff
@@ -9,3 +9,10 @@ type EmbeddingClient interface {
 	CreateEmbedding(ctx context.Context, input string) ([]float32, error)
 	CreateEmbeddingForQuery(ctx context.Context, input string) ([]float32, error)
 }
+
+// BatchEmbeddingClient is an optional provider capability for embedding multiple documents in one
+// request. Workers use it when batching is enabled; providers that do not implement it continue to
+// receive one request per document.
+type BatchEmbeddingClient interface {
+	CreateEmbeddings(ctx context.Context, inputs []string) ([][]float32, error)
+}
```
xernobyl, line 19: [P4] Real but cosmetic. The interface doesn't document the ordering/count contract (exactly one vector per input, in input order). The only current implementation (`openai/client.go`) already enforces and documents this correctly, so this only matters for a future provider implementing `BatchEmbeddingClient` — worth the one-line doc addition CodeRabbit proposed.

BhagyaAmarasinghe:
Also addressed the outside-diff review note in f8bee81: tenant-scoped backfill completion logs now include tenant_id, while global backfills omit the field and retain tenant_scoped=false.

xernobyl (commented):
**no findings** — reviewed at `f8bee813b1`

**0 blocker · 0 should · 0 nit posted (2 low-confidence nits noted below, suppressed per re-review policy) · 0 pre-existing** — reviewed at `f8bee81`

Re-review. This PR was already reviewed at `72de631` (CodeRabbit + a prior house-style pass, both on this PR). `f8bee81` ("fix: address embedding throughput review feedback"), pushed today, addresses every `should`/`nit` from that round:

- **cmd/worker/app.go** — the embedding-batch flush on shutdown now uses its own `embeddingBatchFlushTimeout` (5s) instead of reusing the full `ShutdownTimeout`, so total graceful shutdown (River stop + escalation + this flush) stays inside one grace period instead of reaching ~65s. Fixed as proposed.
- **Dockerfile** — `HEALTHCHECK` now no-ops the HTTP probe unless PID 1 is `hub-api`, so `hub-worker`/`backfill-embeddings` sharing the image no longer inherit a failing API health check. Fixed.
- **internal/config/config_test.go** — the batch-max-wait assertion now uses `37` instead of the `25` default, so it actually exercises the `EMBEDDING_BATCH_MAX_WAIT_MS` env binding rather than passing regardless. Fixed.
- **internal/service/embedding_client.go** — `BatchEmbeddingClient`'s doc comment now states the one-vector-per-input, input-order contract. Fixed.
- **internal/service/embedding_batcher.go** — new `retryIndividually` closes the "one bad input fails every request in the batch" finding: on a batch rejection with more than one active request, each request now gets its own single-item provider call instead of inheriting the batch-wide error, so an unrelated document no longer burns a River retry attempt for a peer's bad input. Verified against `TestBatchingEmbeddingClientProviderErrorRetriesRequestsIndividually` — the bad input still errors, its peers still succeed, and per-request call counts match.

The one Security/Major candidate from the prior round — tenant-boundary handling in `BackfillTenantEmbeddingsWithInputKind` (feedback_records_service.go:647) — remains a verified false positive: `tenant_id` is immutable on `feedback_records`, so a record selected inside tenant A's scoped page can never later belong to tenant B before the enqueued job runs. Re-checked independently against the fix commit: the new `ListTenantFeedbackRecordIDsForBackfillByInputKind` / `CountTenantFeedbackRecordsForBackfillByInputKind` repository methods use the same bind-parameter pattern (`fr.tenant_id = $N`, no string concatenation of the value), and `-tenant-id` is an operator CLI flag on `backfill-embeddings`, not attacker-reachable input. No new gap here.

I hunted the fix commit's genuinely new code — the Dockerfile healthcheck gate and `retryIndividually` — since neither existed at the prior review's head. Two low-confidence points surfaced; both are noted here rather than posted inline, per re-review policy (new findings below `should` stay out of the diff):

1. The Dockerfile's `[ "$(readlink /proc/1/exe)" != "/app/hub-api" ] || wget ...` gate identifies the running role by PID 1's exe path. That holds for today's direct exec-form entrypoints, but would silently stop probing (always "healthy") if `hub-api` is ever started behind a signal-forwarding wrapper such as `tini`/`dumb-init` — worth a one-line note if that ever becomes a deployment pattern; not applicable today.
2. `retryIndividually` fans a failed batch out into `len(active)` immediate, un-backed-off individual retries regardless of failure cause. For an input-specific rejection that's exactly the fix; for a systemic failure (429/5xx/timeout) it briefly doubles load on an already-struggling provider before River's own per-job backoff takes over. Bounded by `EMBEDDING_BATCH_SIZE`/`EMBEDDING_BATCH_MAX_IN_FLIGHT` (both default to no batching), and no worse than the pre-existing non-batch fallback path's exposure to the same provider — not asking for a change, just flagging the trade-off.

**Checked** — tenancy/tenant-scoping (both the pre-existing and newly-added tenant-scoped backfill queries), correctness (shutdown ordering, batch retry semantics, config defaults, decode/validation of provider responses), security (secrets, logging, SQL parameterization, CLI-only inputs), drift (comments vs. behavior in the touched files), and the full diff against `main` plus the incremental fix-commit diff against the prior review's head.
**Skipped** — frontend (no UI in this diff); simplicity (no win beyond what CodeRabbit and the prior round already covered).
**Not verifiable here** — I did not run `make build`/`go test`/the Postgres-backed integration suite myself in this session; the PR's own CI (Unit Tests, Integration Tests pg16/17/18, Coverage, API Contract Tests, Code Quality) is green at this head, matching the author's local run listed in the PR body.

**Checked** — tenancy, correctness, security, drift, tests. **Skipped** — frontend (no UI in the diff); simplicity (no win beyond what CodeRabbit/the prior review round already covered).

**Not verifiable from a static review** — make build / go test / the Postgres-backed integration suite were not re-run in this session — relying on the green CI at this head.

_Reply per finding with the fix commit, a reason it's wrong, or a ticket if you're deferring it — a disputed finding closed with a reason is a good outcome._

xernobyl (commented):
**0 blocker · 0 should · 4 nit** — reviewed at `f8bee813b1`

Third pass on this PR, this time a dedicated security-oriented read of the new code rather than a reconciliation of the prior rounds. Nothing here blocks the merge. The two things I most wanted to break — the batcher's concurrency and the cross-tenant vector-misassignment risk that batching introduces — both hold up:

- **Vector misassignment across tenants.** Coalescing turns "one document per request" into "N tenants' documents in one request", so a provider returning results out of order would write tenant A's vector onto tenant B's record. `decodeEmbeddingResponse` (`internal/openai/client.go:167-206`) defends this properly: exact count match, per-item index range check, duplicate-index rejection, dimension check, and NaN/Inf rejection, with placement by `item.Index` rather than by arrival order. `retryIndividually` routes through the same decoder with `expected = 1`. This is the right defense and it is complete.
- **Batcher concurrency.** I traced the `pending`/`timer`/`active`/`semaphore`/`forceStop` interactions for shutdown races, `WaitGroup` misuse (positive `Add` after `Wait`), semaphore leaks on the `forceStop` path, and goroutine leaks in `providerContext`. All the `Add` calls happen under the same mutex that `Shutdown` uses to drain `pending`, and `closed` blocks new pending, so no positive `Add` can follow the drain. The semaphore is released only on the branch that acquired it, and `cleanup` always closes `providerDone`. No findings.
- **SQL.** Both new tenant-scoped queries bind `tenant_id` as a parameter; the only string concatenation is over compile-time-constant fragments (`eligibility`, `tenantClause`). Placeholder numbering matches the appended args in both the list ($4) and count ($2) paths. No injection surface, and `-tenant-id` is an operator CLI flag regardless.
- **Logging.** The new `tenant_id` attr is a tenant identifier, not content; no record text or provider credentials reach the log lines.

Four nits, all operational rather than correctness. Posting them for the record; none needs to hold the merge.

---

**[P3]** `internal/repository/enrichment_status_repository.go:132-141` — the new `countTaxonomyEmbeddingBacklogAggregateSQL` omits the `fr.field_type = 'text'` filter that `countEnrichmentBacklogAggregateSQL` (directly above it, line 124-125) applies. The live path that produces a taxonomy embedding is `SetTranslation` → `enqueueTaxonomyEmbedding` (`internal/service/feedback_records_service.go:218`), and the translation worker gates on `IsTextField` (`internal/workers/feedback_translation.go:44`). So a non-text record carrying non-empty `value_text` is counted as pending but nothing in the live path will ever drain it — it becomes a permanent floor under the new gauge. Only `backfill-embeddings` clears it, which is consistent with the comment's stated intent, but it does mean the gauge's steady-state floor is not zero.

**[P3]** `cmd/api/app.go:615-620` — the gauge is gated on `embeddingProvider != "" && (Taxonomy.ServiceURL != "" || Taxonomy.ServiceToken != "")`, but not on translation being configured. Since `enqueueTaxonomyEmbedding` fires only from the `SetTranslation` write, a deployment with taxonomy + embeddings but `TRANSLATION_PROVIDER`/`TRANSLATION_MODEL` unset never auto-produces a taxonomy embedding at all — the gauge climbs with ingest and only an operator-run backfill ever brings it down. Every sibling gauge in the same poller is gated on its own enrichment being configured (`translationConfigured`, `sentimentConfigured`, `emotionsConfigured`); this one breaks that symmetry. Since the whole point of the gauge is to be alertable, a version that can't reach zero on a valid configuration is worth either gating on `translationConfigured` too, or documenting as "run the backfill" rather than "the pipeline is stuck".

**[P3]** `internal/service/embedding_batcher.go:249-253` — `retryIndividually` runs inside `dispatch`, so it holds the in-flight semaphore slot (acquired at line 205, released by the deferred receive) across `len(active)` *sequential* provider calls. At the default `EMBEDDING_BATCH_MAX_IN_FLIGHT=1` that means one rejected batch stalls all embedding throughput for N serial round-trips, each carrying the SDK's own timeout budget. The isolation semantics are right — this is purely about the failure path being much slower than the happy path. Worth either fanning the retries out concurrently or letting them bypass the semaphore, though at `BATCH_SIZE=1` (the default) it never fires.

**[P4]** `cmd/backfill-embeddings/main.go:162-181` — `-tenant-id` isn't validated against an existing tenant, so a typo produces `Found 0 eligible missing embedding(s)` / `Enqueued 0 embedding job(s)`. That's indistinguishable from the legitimate "this tenant is fully backfilled" result, which is unfortunate for a flag whose whole purpose is a pre-flight check before an unbounded enqueue. A tenant-exists check, or wording that distinguishes the two, would make `--count-only` trustworthy as a canary.

---

**Checked** — cross-tenant vector assignment under batching, provider-response validation, batcher concurrency (shutdown ordering, WaitGroup, semaphore, goroutine leaks, context propagation), SQL parameterization and placeholder numbering on both new tenant-scoped queries, tenant scoping of the new CLI flags, secrets/PII in the new log lines, config defaults vs. `.env.example`, and comment-vs-behavior drift in the touched files.
**Skipped** — frontend (no UI in this diff).
**Not verifiable here** — I did not run `make build`, `go test`, or the Postgres-backed integration suite in this session; CI is green at this head across Unit, Integration pg16/17/18, Coverage, API Contract, and Code Quality.

BhagyaAmarasinghe:
Addressed all four operational review nits in `c7c5760`:

- taxonomy backlog counting now includes only `field_type = 'text'`, with an integration assertion that categorical `value_text` does not raise the gauge;
- the taxonomy-embedding gauge is now enabled only when embeddings, translation, and taxonomy are all configured;
- failed-batch single-item retries run concurrently, so the batch slot is held for one additional provider round trip rather than N sequential calls;
- tenant-scoped backfills now validate the tenant against Hub's available tenant data (`feedback_records` or `tenant_settings`) and fail clearly for an unknown ID instead of reporting a misleading zero.

Validated with `make build`, lint at 0 issues, targeted race tests, and the full integration suite against a fresh PostgreSQL 18 + pgvector database.

xernobyl (commented):
**no findings** — reviewed at `c7c5760ae7`

All four nits from my previous pass are resolved in `c7c5760`. Verified each against the code rather than the commit message; every one came with a test that would actually catch a regression, which is the part I care about more than the fix itself.

- **`field_type = 'text'` on the taxonomy gauge** (`enrichment_status_repository.go:127-137`) — added, and the comment now says *why* it's load-bearing rather than just that it's there. The integration test earns its keep: it takes a baseline count, seeds one text and one categorical record, and asserts the gauge moved by exactly `+1`. Using a before/after delta instead of an absolute means it stays honest regardless of what other rows the suite leaves behind, and the categorical row would have shown up as `+2` under the old predicate.
- **Gauge gating** (`cmd/api/app.go:665-675`) — `taxonomyEmbeddingBacklogModel` now requires embeddings **and** translation **and** taxonomy, which closes the case where the gauge could never reach zero because nothing produced taxonomy embeddings in the first place. `TestTaxonomyEmbeddingBacklogModelRequiresCompletePipeline` covers the complete config plus each of the four ways to break it.
- **`retryIndividually` concurrency** (`embedding_batcher.go:269-286`) — the retries now run under a `WaitGroup` instead of serially, so a rejected batch holds its in-flight slot for one additional round trip rather than N. This does mean a failed batch briefly issues up to `BatchSize` concurrent single calls from inside one `MaxInFlight` slot, which is the trade I was asking for and is still bounded by `EMBEDDING_BATCH_SIZE`. The test change is the good part: all three single calls are blocked on a channel and the test requires all three to have *started* within a 1s deadline before unblocking them — a sequential implementation deadlocks and fails instead of quietly passing. Loop-variable capture is fine on the repo's Go version.
- **Tenant validation** (`backfill-embeddings/main.go:161-175`, `embeddings_repository.go:374-395`) — `TenantExistsForEmbeddingBackfill` checks both `feedback_records` and `tenant_settings`, which is the right call given Hub has no authoritative tenants table, and the doc comment says so. Bound parameter, no concatenation. It runs *before* the `-count-only` branch, so the canary path is covered too — that was the specific gap, since `--count-only` is documented as the pre-flight before an unbounded enqueue and "Found 0" was previously indistinguishable from a typo. A genuinely new tenant with no data yet now fails rather than reporting zero, which is the correct reading: there is nothing to back-fill.

Nothing new surfaced in the fix commit. The batching hot path, the provider-response validation, and the tenant-scoped SQL are unchanged from the pass where I went at them properly, and this commit only touches the gauge predicate, the gauge's config gate, the retry fan-out, and the new tenant pre-check.

CI is green at this head across all 11 checks, including Integration Tests on pg16/17/18 — which is where the two new assertions actually run. No blockers from me.

xernobyl (approved):
Approving at `c7c5760ae7`. All four nits from my pass are fixed and verified in the review above, and nothing new surfaced in the fix commit.

Recapping what I actually leaned on, since this PR turns "one document per provider request" into "N tenants' documents per request" and that's the part worth being sure about:

- **Vector misassignment can't happen silently.** `decodeEmbeddingResponse` enforces exact count, per-item index range, duplicate-index rejection, dimension match, and NaN/Inf rejection, and places results by `item.Index` rather than arrival order. `retryIndividually` goes through the same decoder with `expected = 1`. The `BatchEmbeddingClient` interface now states the one-vector-per-input, input-order contract, so a future provider implementation has something to be held to.
- **Batcher concurrency holds.** Every `active.Add` happens under the mutex `Shutdown` uses to drain `pending`, and `closed` blocks new pending, so no positive `Add` can follow the drain. The semaphore is released only on the branch that acquired it, and `providerContext`'s cleanup always closes `providerDone`.
- **Blast radius is small.** `EMBEDDING_BATCH_SIZE` defaults to `1`, so the batcher isn't constructed at all until an operator opts in, and the non-batch path is untouched.

The taxonomy gauge is the piece I'd watch first in production: it now correctly counts only text records and only publishes when embeddings, translation, and taxonomy are all configured, so a non-zero reading with an empty embedding queue genuinely means "run the backfill" rather than "this metric never drains". That's the property that makes it alertable, and it's the one that was worth the two extra rounds.

Nice work on the tests in `c7c5760` — the before/after delta on the gauge count and the started-within-deadline construction on the concurrent retry both fail loudly on a regression instead of quietly passing.

PR #123, opened 2026-08-17
feat: add as_of and disabled_reason to the enrichment status endpoint (ENG-2375)

xernobyl:
## What does this PR do?

First of several PRs on [ENG-2375](https://linear.app/formbricks/issue/ENG-2375/extend-status-endpoint-with-more-info). Adds two fields to `GET /v1/enrichment-status` that need no migration and no new query.

`enabled: false` was unactionable on its own. Three very different situations collapsed into it — the deployment has no provider configured, the tenant switched the enrichment off, or translation has no resolvable target language — each with a different fix and a different person who can apply it. The UI could only hide the enrichment rather than explain it.

- **`disabled_reason`** — `not_configured` | `switched_off` | `no_target_language`. The service already evaluated those gates separately and then collapsed them into one bool, so this is mostly a matter of not throwing the answer away. `enabled` and `disabled_reason` are derived from the same value, so a response cannot carry an enabled enrichment with a reason or a disabled one without.
- **`as_of`** — stamped when the counts are computed, not when the response is serialized. The endpoint is polled, so two responses are enough for a client to derive throughput and an ETA from the change in `done` over the change in `as_of`. Hub computes neither.

When both the deployment gate and the tenant switch are closed, the deployment gate is reported. It is the outer gate, and pointing a tenant at a switch that changes nothing while no provider exists is the wrong fix.

### Open question for the consumer — @Dhruwang

The ticket specified `disabled_reason: … | null`. This PR emits it as **absent** when an enrichment is enabled, not `null`.

I originally justified that with "the spec has no nullable fields" — **that was wrong**, and I want to correct it here rather than let it stand. `WebhookPublicData.disabled_reason` and `WebhookData.disabled_reason` both use `type: [string, "null"]`, the OpenAPI 3.1 nullable form; my earlier check grepped for the 3.0 `nullable: true` spelling and found nothing.

So the honest position is that both forms exist in this spec, your ticket asked for `null`, and the sibling field of the same name is nullable — which makes this one the odd one out. Switching costs a `*DisabledReason` and `type: [string, "null"]`, roughly six lines including the test. Happy to change it before this merges; it is much cheaper now than after later PRs build on the response shape.

Worth flagging separately: `disabled_reason` now means two unrelated things in this spec — why a *webhook* was disabled (free-form, nullable) versus which gate closed an *enrichment* (enum, optional). Stainless generates per-schema models so there is no type clash, but it is a readability trap.

### Before / after

Captured by running `origin/main` (`01f6e8b`) and this branch side by side against the **same database and the same tenant**, so only the code differs.

**State A — tenant has no target language, deployment has no default**

```jsonc
// before
{"tenant_id":"demo-t1","translation":{"enabled":false,"eligible":0,"done":0},"sentiment":{"enabled":true,"eligible":0,"done":0},"emotions":{"enabled":true,"eligible":0,"done":0}}

// after
{"tenant_id":"demo-t1","as_of":"2026-08-17T15:47:32.581756Z","translation":{"enabled":false,"eligible":0,"done":0,"disabled_reason":"no_target_language"},"sentiment":{"enabled":true,"eligible":0,"done":0},"emotions":{"enabled":true,"eligible":0,"done":0}}
```

**State B — target set to `de-DE`, tenant switched sentiment off**

```jsonc
// before
{"tenant_id":"demo-t1","translation":{"enabled":true,"eligible":0,"done":0},"sentiment":{"enabled":false,"eligible":0,"done":0},"emotions":{"enabled":true,"eligible":0,"done":0}}

// after
{"tenant_id":"demo-t1","as_of":"2026-08-17T15:47:32.635006Z","translation":{"enabled":true,"eligible":0,"done":0},"sentiment":{"enabled":false,"eligible":0,"done":0,"disabled_reason":"switched_off"},"emotions":{"enabled":true,"eligible":0,"done":0}}
```

**State C — no providers configured at all; tenant state unchanged from B**

```jsonc
// before
{"tenant_id":"demo-t1","translation":{"enabled":false,"eligible":0,"done":0},"sentiment":{"enabled":false,"eligible":0,"done":0},"emotions":{"enabled":false,"eligible":0,"done":0}}

// after
{"tenant_id":"demo-t1","as_of":"2026-08-17T15:48:11.503321Z","translation":{"enabled":false,"eligible":0,"done":0,"disabled_reason":"not_configured"},"sentiment":{"enabled":false,"eligible":0,"done":0,"disabled_reason":"not_configured"},"emotions":{"enabled":false,"eligible":0,"done":0,"disabled_reason":"not_configured"}}
```

**The whole point, in one comparison.** `sentiment` in states B and C, on `main` today:

```json
B: {"enabled": false, "eligible": 0, "done": 0}
C: {"enabled": false, "eligible": 0, "done": 0}
```

Byte-identical, and they mean different things: in B the tenant turned sentiment off and can turn it back on; in C no provider is configured and nothing the tenant does will help. After this PR they read `switched_off` and `not_configured`.

Note also that state C's tenant still has `sentiment_enabled: false` *and* a target language set, yet everything reports `not_configured` — the deployment gate winning over the tenant switch, as intended.

## How should this be tested?

Unit tests cover the gate matrix; the integration suite and a manual pass cover the wire format.

**Automated**

- `make test-unit` — the service table now pins every gate combination, including the two precedence cases (deployment gate beats the tenant switch; the target check does not shadow the deployment gate for translation).
- `make tests` — integration suite. Requires River's migrations as well as goose's, or `TestTenantTranslationBackfillWorker_Work` fails with `relation "river_job" does not exist`:

```
goose -dir migrations postgres "$DATABASE_URL" up
river migrate-up --database-url "$DATABASE_URL"
DATABASE_URL=... API_KEY=... make tests
```

**Manual — the part unit tests structurally cannot check**

The previous HTTP test unmarshalled the response back into the same Go struct and compared structs, which is symmetric with the marshal that produced it and therefore blind to the wire format. There is now a `wire format matches the published schema` test asserting the JSON by field name, and it was verified by deleting `omitempty` from `DisabledReason` — it fails on exactly that line, and passes again when restored.

To reproduce end to end:

1. Run the API with all three enrichments "configured". No real provider calls are made — the status endpoint only reads whether provider+model are set:

```
DATABASE_URL=... API_KEY=k PORT=8099 \
  SENTIMENT_PROVIDER=openai SENTIMENT_MODEL=gpt-4o-mini \
  EMOTIONS_PROVIDER=openai EMOTIONS_MODEL=gpt-4o-mini \
  TRANSLATION_PROVIDER=openai TRANSLATION_MODEL=gpt-4o-mini \
  TRANSLATION_DEFAULT_LANGUAGE="" go run ./cmd/api
```

2. `curl -H "Authorization: Bearer k" "localhost:8099/v1/enrichment-status?tenant_id=t1"` → translation `no_target_language`, others enabled with **no** `disabled_reason` key at all.
3. Set a target and switch sentiment off, then re-poll → translation enabled, sentiment `switched_off`:

```
curl -X PUT -H "Authorization: Bearer k" -H 'Content-Type: application/json' \
  -d '{"target_language":"de-DE","sentiment_enabled":false}' \
  localhost:8099/v1/tenants/t1/settings
```

4. Restart with no `*_PROVIDER` / `*_MODEL` vars → all three `not_configured`, even though the tenant state from step 3 is unchanged.

I ran 1–4 against a scratch database and additionally validated the live responses against the enum and required-list parsed out of `openapi.yaml`: required fields present, `as_of` parses as RFC 3339, no enabled enrichment carries a reason, no disabled one lacks it, nothing outside the enum, no empty strings.

### Required

- [x] Ran `make build`
- [x] Merged the latest changes from main onto my branch — cut from `01f6e8b`, 0 commits behind
- [x] If database schema changed: n/a, no migration in this PR

### Appreciated

- [ ] Updated docs in `docs/` if changes were necessary — n/a, this repo has no `docs/`
- [ ] Ran `make tests-coverage` for meaningful logic changes — not run; the logic here is a gate resolver fully covered by the service table

### DIFF
--- a/internal/api/handlers/enrichment_status_handler_test.go
+++ b/internal/api/handlers/enrichment_status_handler_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -49,9 +50,12 @@ func TestEnrichmentStatusHandler_GetStatus(t *testing.T) {
 	t.Run("returns 200 with the status body", func(t *testing.T) {
 		want := &models.EnrichmentStatusResponse{
 			TenantID:    "t1",
+			AsOf:        time.Date(2026, 8, 17, 12, 0, 0, 0, time.UTC),
 			Translation: models.EnrichmentTypeStatus{Enabled: true, Eligible: 10, Done: 4},
 			Sentiment:   models.EnrichmentTypeStatus{Enabled: true, Eligible: 8, Done: 8},
-			Emotions:    models.EnrichmentTypeStatus{Enabled: false},
+			Emotions: models.EnrichmentTypeStatus{
+				Enabled: false, DisabledReason: models.DisabledReasonNotConfigured,
+			},
 		}
 		h := NewEnrichmentStatusHandler(&fakeEnrichmentStatusService{resp: want})
 		rec := httptest.NewRecorder()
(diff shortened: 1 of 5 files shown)


xernobyl:
Closing this — not abandoning it. The work is being consolidated into a single PR covering the whole of ENG-2375 (status fields, terminal failure taxonomy, reconciler, retry API, metrics), because splitting it into a stack would spend more time in the review queue than it saves. Same branch, `feat/ENG-2375_enrichment-status-fields`; a new PR goes up when the whole thing is ready.

The two commits here are unchanged and carry forward. The open question about `disabled_reason` being absent versus `null` moves to ENG-2375 so it doesn't get buried with this PR.

PR #125, opened 2026-08-19
feat: report why an enrichment is off and what failed, durably (ENG-2375)

xernobyl:
## What does this PR do?

`GET /v1/enrichment-status` returned `{enabled, eligible, done}`, and the consumer derived "in progress" as `eligible - done`. That number quietly merges three unrelated populations: work that is genuinely queued, records that failed permanently, and records that were never enqueued at all. Dhru's repro on ENG-2375 is the third one — create a record with AI credentials unset, set them, create a second record, and the UI reports 1 of 2 with the first row spinning forever.

This PR makes the difference reportable. It does **not** yet make it self-correcting — the reconciler and the retry endpoint are ENG-2376, stacked on top of this.

**Phase 1 fields.** `as_of` (when the counts were taken, so a polling client can derive throughput and an ETA without us computing either) and `disabled_reason` — `not_configured` / `switched_off` / `no_target_language`. The service already evaluated those three gates separately and then collapsed them into one bool; `enabled: false` on its own is unactionable, since an operator who never configured a provider and a tenant who turned the enrichment off need completely different fixes.

**A terminal-vs-transient taxonomy.** `huberrors.TerminalProviderError` with a closed reason set (`content_filter`, `refusal`, `length`, `recitation`). Both provider clients already detected these and threw the distinction away. Two bugs fell out of building it:

- a `length` finish reason was being read *after* the content, so a truncated translation was stored as a finished one and a truncated JSON body read as a transient parse glitch. It is permanent for that text at that model, and it is now checked before the content is touched.
- `resp.Candidates[0]` in the Gemini path panics on `"candidates": [null]` — the SDK's own `Text()` has the same gap (it guards the slice length and a nil `Content`, but not a nil candidate).

**Durable failure state** — migration `022`, `feedback_record_enrichment_failures`, one row per (record, enrichment) with whether retrying could ever help. It has to be durable because River reaps discarded jobs (7d default; we override only `CompletedJobRetentionPeriod`), so a queue-derived failure count silently drains to zero. Rows are **advisory**: `failed` is reported as "a row exists AND the record is still un-enriched", so a later success needs no cleanup write and the enrichment hot path stays free of bookkeeping. `reason` is a bounded enum, never the provider's error string — refusal text is model-generated but routinely paraphrases the input, so storing it would put paraphrased customer feedback at rest and then into the API.

**The counts** — `failed` and `failed_terminal`, from the same scan the endpoint already did (three PK-lookup joins, no second query). The failure joins are deliberately kept off the cross-tenant aggregate, which is the one documented full-table scan and does not read those columns.

**Cost metrics**, per the OTel GenAI semantic conventions: `gen_ai.client.token.usage` and `gen_ai.client.operation.duration`. Both clients were decoding the `usage` block every provider returns and dropping it, so there was no way to answer what enrichment costs. Plus `hub_enrichment_terminal_total{enrichment,reason}` and `hub_enrichment_failed_records{enrichment,terminal}`. `tenant_id` is deliberately absent from all of them, and GenAI *content* attributes are prohibited in code and pinned by a test — `gen_ai.input.messages` carries the prompt, which is customer feedback text.

A terminal failure now returns `river.JobCancel` rather than burning the remaining attempts, which is worth real money: a re-sweep of a permanently-failing corpus costs one provider call per record instead of `MAX_ATTEMPTS` (measured below).

**Measured, because the failure columns are not free.** The three marker joins originally carried no tenant-side predicate — the tenant boundary must be `feedback_records.tenant_id`, never the denormalized stamp — which left the planner nothing to filter the marker table by, so it hash-joined the whole thing three times. On 1M records with 300k markers, a tenant holding 2,500 records and *no failures of its own* went from 3ms to 42ms, scaling with the **global** marker count rather than its own (7ms at 10k markers, 18ms at 100k, 32ms at 300k). One tenant's provider outage taxing everyone else's status endpoint, on something the UI polls every five seconds per open page.

Each join now also keys on the marker's `tenant_id`, which restores `idx_enrichment_failures_tenant_enrichment` and puts that query back to 3.9ms — the plan switches from three sequential scans to three index scans. **Additional, not instead of:** `fr.tenant_id` stays in the outer WHERE and stays the boundary, and a join predicate can only narrow which marker attaches to a record, so it cannot reach another tenant's row. The trade is that a marker whose stamp disagreed with its record would stop counting; nothing can produce one, and under-counting is the safe direction. The fixture that covered it now asserts the half that matters — the stamped tenant, owning no record, sees nothing.

The two leader-tick aggregates were measured too and are fine: the backlog scan 380–450ms and the new `CountFailedRecordsAggregate` 118–135ms, both every five minutes on one replica under a statement timeout.

Rebased onto `main` after #122 landed. The records-scoped purge it added removes these markers by `ON DELETE CASCADE` — that is the only derived table it leaves to a cascade, and a cascade reports no row count, so nothing would notice if it stopped firing. There is now a test for it.

Linear: https://linear.app/formbricks/issue/ENG-2375/extend-status-endpoint-with-more-info
Follow-up (stacked on this): https://linear.app/formbricks/issue/ENG-2376/auto-re-queue-failed-jobs

### API change

```
GET /v1/enrichment-status?tenant_id=t1
```
```json
{
  "tenant_id": "t1",
  "as_of": "2026-08-19T16:17:24Z",
  "translation": { "enabled": true,  "eligible": 2000, "done": 1852, "failed": 0, "failed_terminal": 148 },
  "sentiment":   { "enabled": true,  "eligible": 2000, "done": 1830, "failed": 3, "failed_terminal": 167 },
  "emotions":    { "enabled": false, "disabled_reason": "switched_off", "eligible": 0, "done": 0, "failed": 0, "failed_terminal": 0 }
}
```

`enabled` is unchanged and still means what it did. `disabled_reason` is present exactly when `enabled` is false — the two are derived from one decision, so they cannot contradict each other. A disabled enrichment reports zeroes for everything, failures included: the API must not show a backlog of failed work for a pipeline that will never run.

## How should this be tested?

⚠️ **If you ran `goose up` on this branch before `e2b2f83`, recreate your database.** Migration `022`'s CHECK gained a `write_failed` reason after the file had already been applied, and goose tracks by version, so `goose up` is a no-op and your constraint is the old one. `dropdb && createdb && goose up`, or re-apply the two `ALTER TABLE ... reason_valid` statements by hand.

**The suites** (`DATABASE_URL` pointing at a fresh database with pgvector, `goose up`, `river migrate-up`):

```
make build
make lint
go test ./internal/... ./cmd/...     # 16 packages
make tests                            # integration, real Postgres
make lint-openapi
make migrate-validate
```

**The interesting bits are the fault injections**, since a green suite proves very little on its own. Each of these should turn the suite red, and does:

- gate the failure counts on the markers' own `tenant_id` instead of the record's (the "optimization" migration 022 warns about) → `TestCountEnrichmentStatusCountsFailures`
- drop the marker table's foreign key → `TestEnrichmentFailuresPurgedWithTenant` and `TestPurgeFeedbackRecordsByTenant`
- drop either CHECK constraint → `TestEnrichmentFailureMarkerConstraints`
- add a `tenant_id` or `gen_ai.input.messages` attribute to any metric → `TestGenAIMetricsCarryOnlyAllowedAttributes`
- remove the marker write from the persist path → `TestFeedbackSentimentWorker_WriteFailureRecordsMarker`

**End to end.** I ran the whole thing against a synthetic OpenAI-compatible provider that injects a configurable mix of 429s, 500s, refusals, content-filter blocks and truncations — 3200 records across five tenants, ~14k provider calls, 916 assertions, 0 failures:

- the API's counts agree with an **independent** SQL computation (not the query under test) for every tenant and enrichment
- `eligible = done + failed + failed_terminal` closes exactly, with nothing stranded, and `done + failed + failed_terminal` never went backwards across ~90 live samples taken while work was in flight
- all three terminal reasons came through the real classification path — refusal 359, content_filter 239, length 235 — and 747 of 841 terminal give-ups happened on the first call
- re-sweeping 677 permanently-failing records cost **exactly 677 provider calls**, not 3×; for contrast a total outage over 200 records spent 1803 (200 × 3 enrichments × 3 attempts) and then recovered to zero failures with every marker still in the table, unread
- every stored translation was the provider's complete output — the truncation fix, asserted end to end
- 500,808 input / 143,088 output tokens recorded, and no tenant id, feedback text, or unbounded value on any exported metric label

**What is not covered**, so you know where to push back: the smoke run used `MAX_CONCURRENT=25` per pipeline — 75 concurrent jobs — in a single `hub-worker` against a single `hub-api`. Two live API replicas competing for leadership was never exercised end to end, which is exactly what the backlog/failed-records poller's leader election and its withdraw-on-handover behaviour exist for. Only the unit test covers withdrawal, in isolation.

**By hand**, if you want Dhru's original repro: create a record with `SENTIMENT_PROVIDER`/`SENTIMENT_MODEL` unset, set them, create a second record, and hit the endpoint. Before this PR the first record is invisible in `eligible - done`; now the tenant reads `not_configured` while it is off, and the stranded record shows up as neither done nor failed — which is honest, and is what ENG-2376 then drains.

### Required

- [x] Ran `make build`

### Appreciated

### DIFF
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -816,10 +816,16 @@ func runEnrichmentBacklogPoller(
 		// backlog gauge and no second election is needed. Its query is cheap next to the backlog
 		// scan: it reads the markers, of which there are few, not every feedback record.
 		//
-		// queryCtx, not ctx — this shares the tick's timeout budget. Handing it the process-lifetime
-		// context would let a slow count pin a pool connection until shutdown, which is the exact
-		// thing enrichmentBacklogQueryTimeout exists to prevent.
-		refreshFailedRecords(queryCtx, statusRepo, failures)
+		// Its OWN deadline, not the leftover of queryCtx. Both need a bound — an unbounded count
+		// can pin a pool connection until shutdown, which is what enrichmentBacklogQueryTimeout
+		// exists to prevent — but sharing one budget couples them the wrong way round: the backlog
+		// scan above is the documented whole-table sequential scan, so on a large deployment it
+		// eats most of the window, and this count then times out on every tick and withdraws the
+		// gauge for as long as the scan stays slow. A separate timeout off the same parent keeps
+		// the bound and drops the coupling.
+		failedCtx, cancelFailed := context.WithTimeout(ctx, enrichmentBacklogQueryTimeout)
+		refreshFailedRecords(failedCtx, statusRepo, failures)
+		cancelFailed()
 	}
 
 	update()
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -815,7 +815,11 @@ func runEnrichmentBacklogPoller(
 		// Refreshed by the SAME leader on the SAME tick, so the failure gauge cannot drift from the
 		// backlog gauge and no second election is needed. Its query is cheap next to the backlog
 		// scan: it reads the markers, of which there are few, not every feedback record.
-		refreshFailedRecords(ctx, statusRepo, failures)
+		//
+		// queryCtx, not ctx — this shares the tick's timeout budget. Handing it the process-lifetime
+		// context would let a slow count pin a pool connection until shutdown, which is the exact
+		// thing enrichmentBacklogQueryTimeout exists to prevent.
+		refreshFailedRecords(queryCtx, statusRepo, failures)
 	}
 
 	update()
@@ -1082,10 +1086,17 @@ func (a *App) awaitEnrichmentBacklogPoller(ctx context.Context) {
 	}
 }
 
-// refreshFailedRecords republishes the cross-tenant failed-record gauge. Best effort: a failure
-// here logs and leaves the previous reading rather than tearing down the backlog gauge with it,
-// because the two answer different questions and one being unavailable is not a reason to lose
-// both.
+// refreshFailedRecords republishes the cross-tenant failed-record gauge.
+//
+// A failed query WITHDRAWS this gauge rather than leaving its last reading published, following
+// the rule EnrichmentBacklogMetrics states outright: exporting nothing is the honest state,
+// because absence is visible and a stale value is not. The failure mode it avoids is specific —
+// the backlog scan succeeds so this process keeps leadership, this query times out, and a
+// dashboard would otherwise show a plausible, unchanging failure count that no longer reflects
+// the database, with nothing absent to alert on.
+//
+// It withdraws only ITS OWN series. The backlog gauge is refreshed by a query that succeeded and
+// has no reason to be torn down alongside.
 func refreshFailedRecords(
 	ctx context.Context,
 	statusRepo *repository.EnrichmentStatusRepository,
@@ -1097,8 +1108,12 @@ func refreshFailedRecords(
 
 	counts, err := statusRepo.CountFailedRecordsAggregate(ctx)
 	if err != nil {
+		failures.ClearFailedRecords()
+
+		// Shutdown cancels the query mid-flight; that is not a poll failure worth logging, for the
+		// same reason the backlog poller skips it.
 		if ctx.Err() == nil {
-			slog.WarnContext(ctx, "enrichment failed-records poll failed", "error", err)
+			slog.WarnContext(ctx, "enrichment failed-records poll failed; gauge withdrawn", "error", err)
 		}
 
 		return
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -650,15 +650,16 @@ func (a *App) Run(ctx context.Context) error {
 		go func() {
 			defer close(done)
 
-			runEnrichmentBacklogPoller(ctx, a.db, a.metrics.EnrichmentBacklog, enrichmentBacklogPollConfig{
-				// Trim to stay consistent with NewEnrichmentStatusService (config already canonicalizes
-				// this, so it's defensive symmetry) — the endpoint and the gauge resolve the same target.
-				defaultLang:            strings.TrimSpace(a.cfg.Translation.DefaultLanguage),
-				translationConfigured:  translationConfigured,
-				sentimentConfigured:    a.cfg.Sentiment.Enabled(),
-				emotionsConfigured:     a.cfg.Emotions.Enabled(),
-				taxonomyEmbeddingModel: taxonomyEmbeddingModel,
-			})
+			runEnrichmentBacklogPoller(ctx, a.db, a.metrics.EnrichmentBacklog, a.metrics.EnrichmentFailures,
+				enrichmentBacklogPollConfig{
+					// Trim to stay consistent with NewEnrichmentStatusService (config already canonicalizes
+					// this, so it's defensive symmetry) — the endpoint and the gauge resolve the same target.
+					defaultLang:            strings.TrimSpace(a.cfg.Translation.DefaultLanguage),
+					translationConfigured:  translationConfigured,
+					sentimentConfigured:    a.cfg.Sentiment.Enabled(),
+					emotionsConfigured:     a.cfg.Emotions.Enabled(),
+					taxonomyEmbeddingModel: taxonomyEmbeddingModel,
+				})
 		}()
 	}
 
@@ -725,8 +726,11 @@ func runEnrichmentBacklogPoller(
 	ctx context.Context,
 	db *pgxpool.Pool,
 	backlog observability.EnrichmentBacklogMetrics,
+	failures observability.EnrichmentFailureMetrics,
 	cfg enrichmentBacklogPollConfig,
 ) {
+	statusRepo := repository.NewEnrichmentStatusRepository(db)
+
 	leader := repository.NewEnrichmentBacklogLeader(db)
 	defer leader.Close(ctx)
 
@@ -735,6 +739,7 @@ func runEnrichmentBacklogPoller(
 	// final collect-and-export on shutdown would publish one last reading for a backlog this
 	// process no longer owns.
 	defer backlog.ClearEnrichmentPending()
+	defer clearFailedRecords(failures)
 
 	ticker := time.NewTicker(enrichmentBacklogInterval)
 	defer ticker.Stop()
@@ -760,6 +765,7 @@ func runEnrichmentBacklogPoller(
 			// A failed scan also costs this process its leadership, so withdraw the series rather
 			// than leave it frozen at the last good reading while the new leader publishes its own.
 			backlog.ClearEnrichmentPending()
+			clearFailedRecords(failures)
 
 			// Count the failure so the gap is alertable, then escalate the log from warn to error
 			// once failures persist: a single blip is noise, but a run of them means nobody is
@@ -784,6 +790,7 @@ func runEnrichmentBacklogPoller(
 			// anything this process exported while it was previously the leader, so a handover
 			// leaves exactly one series rather than a live one plus a frozen one.
 			backlog.ClearEnrichmentPending()
+			clearFailedRecords(failures)
 
 			return
 		}
@@ -804,6 +811,11 @@ func runEnrichmentBacklogPoller(
 			backlog.SetEnrichmentPending(
 				observability.EnrichmentTypeTaxonomyEmbedding, counts.TaxonomyEmbeddingPending)
 		}
+
+		// Refreshed by the SAME leader on the SAME tick, so the failure gauge cannot drift from the
+		// backlog gauge and no second election is needed. Its query is cheap next to the backlog
+		// scan: it reads the markers, of which there are few, not every feedback record.
+		refreshFailedRecords(ctx, statusRepo, failures)
 	}
 
 	update()
@@ -1067,5 +1079,52 @@ func (a *App) awaitEnrichmentBacklogPoller(ctx context.Context) {
 	case <-ctx.Done():
 		slog.Warn("enrichment backlog poller did not finish before the shutdown deadline; " +
 			"its final gauge reading may still be exported")
+	}
+}
+
+// refreshFailedRecords republishes the cross-tenant failed-record gauge. Best effort: a failure
+// here logs and leaves the previous reading rather than tearing down the backlog gauge with it,
+// because the two answer different questions and one being unavailable is not a reason to lose
+// both.
+func refreshFailedRecords(
+	ctx context.Context,
+	statusRepo *repository.EnrichmentStatusRepository,
+	failures observability.EnrichmentFailureMetrics,
+) {
+	if failures == nil {
+		return
+	}
+
+	counts, err := statusRepo.CountFailedRecordsAggregate(ctx)
+	if err != nil {
+		if ctx.Err() == nil {
+			slog.WarnContext(ctx, "enrichment failed-records poll failed", "error", err)
+		}
+
+		return
+	}
+
+	// Zero the known buckets before applying, so an enrichment whose last failure was resolved
+	// reports 0 rather than keeping its final non-zero reading forever. The query returns no row
+	// for an empty bucket, which would otherwise be indistinguishable from "not refreshed".
+	for _, enrichment := range []string{
+		observability.EnrichmentTypeSentiment,
+		observability.EnrichmentTypeEmotions,
+		observability.EnrichmentTypeTranslation,
+	} {
+		failures.SetFailedRecords(enrichment, true, 0)
+		failures.SetFailedRecords(enrichment, false, 0)
+	}
+
+	for _, count := range counts {
+		failures.SetFailedRecords(count.Enrichment, count.Terminal, count.Count)
+	}
+}
+
+// clearFailedRecords withdraws the gauge, for the same reason the backlog gauge is withdrawn: a
+// process that is no longer the leader must export nothing rather than a frozen last reading.
+func clearFailedRecords(failures observability.EnrichmentFailureMetrics) {
+	if failures != nil {
+		failures.ClearFailedRecords()
 	}
 }
--- a/cmd/api/app.go
+++ b/cmd/api/app.go
@@ -25,6 +25,7 @@ import (
 	"github.com/formbricks/hub/internal/api/middleware"
 	"github.com/formbricks/hub/internal/api/routes"
 	"github.com/formbricks/hub/internal/config"
+	"github.com/formbricks/hub/internal/llm"
 	"github.com/formbricks/hub/internal/models"
 	"github.com/formbricks/hub/internal/observability"
 	"github.com/formbricks/hub/internal/repository"
@@ -105,6 +106,20 @@ func setupEmbeddingSearchHandler(
 	metrics *observability.Metrics,
 	meterProvider *sdkmetric.MeterProvider,
 ) (*handlers.SearchHandler, error) {
+	// The API works no enrichment jobs, but it does embed SEARCH QUERIES, and those are real
+	// provider calls with real cost. Recording them here means the token metric covers everything
+	// the deployment actually spends rather than only the worker's share.
+	var genAIUsage llm.UsageRecorder
+
+	if meterProvider != nil {
+		recorder, err := observability.NewGenAIMetrics(meterProvider.Meter("hub"))
+		if err != nil {
+			return nil, fmt.Errorf("create gen_ai metrics: %w", err)
+		}
+
+		genAIUsage = recorder
+	}
+
 	embeddingCfg := service.EmbeddingClientConfig{
 		Provider:            embeddingProviderName,
 		ProviderAPIKey:      cfg.Embedding.ProviderAPIKey,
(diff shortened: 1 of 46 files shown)


internal/service/enrichment_client_factory.go
```diff
@@ -29,6 +31,9 @@ type EnrichmentClientConfig struct {
 	BaseURL             string
 	GoogleCloudProject  string
 	GoogleCloudLocation string
+	// UsageRecorder receives each provider call's token counts and duration. nil disables
+	// recording; the backfill commands and tests leave it unset.
+	UsageRecorder llm.UsageRecorder
```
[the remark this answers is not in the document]
xernobyl: Good catch, and the more consequential of the three — fixed in 7c1cfcf.

`internal/googleai` now takes the same `WithUsageRecorder` option and all eight construction sites pass it. Two provider quirks worth knowing about:

- output tokens include `ThoughtsTokenCount`. Thinking tokens bill as output, and this client asks for a zero thinking budget but falls back to the model's default when the model rejects that, so leaving them out would under-report exactly the models where the cost is largest.
- embeddings report no tokens at all — Vertex returns a billable *character* count and AI Studio returns nothing — so those record duration only. The recorder skips zero counts rather than emitting them.

The two constructors also stamp different `gen_ai.provider.name` values, since the API-key path is AI Studio (`gcp.gemini`) and the ADC path is Vertex (`gcp.vertex_ai`) and they bill separately.

Since the failure here was silent — builds, enriches, passes everything, reports nothing — I added an AST parity test (`client_factory_usage_parity_test.go`) that fails if any provider client is constructed without a recorder. Checked it by removing the recorder from one Google factory again; it names the file and line.

internal/workers/enrichment_worker.go
```diff
@@ -252,11 +283,29 @@ func (w *enrichmentWorker[A, R]) handleClassifyError(
...
+			cfg.failureMetrics.RecordTerminalFailure(ctx, cfg.name, string(reason))
+		}
+
+		w.recordOutcome(ctx, "failed_final", start)
+		log.Error(cfg.name+": provider failed permanently for this record, not retrying",
+			"reason", string(reason), "attempt", job.Attempt, "error", err)
+
+		//nolint:wrapcheck // river sentinel: JobCancel must be returned for River to stop retrying
+		return river.JobCancel(fmt.Errorf("%s (terminal, %s): %w", cfg.classifyErrVerb, reason, err))
+	}
```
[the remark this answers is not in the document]
xernobyl: Fair, fixed in 7c1cfcf — on all three give-up paths, not just this one.

The reason it matters more than the millisecond count: `markFailed` blocks for up to a second, so a slow *database* was inflating the enrichment duration histogram, which would make a database incident read as a slow provider. That is the one confusion these metrics were added to prevent, so it is worth the reorder.

internal/workers/feedback_sentiment_test.go
```diff
@@ -416,3 +418,217 @@ func TestFeedbackSentimentWorker_SettingsReadErrorRetriesThenFailsFinal(t *testi
...
+	require.ErrorAs(t, err, &cancelErr,
+		"a terminal failure must cancel the job, not return a plain error River would retry")
+
+	require.Len(t, failures.calls, 1, "the failure must be recorded")
+	assert.True(t, failures.calls[0].Terminal)
+	assert.Equal(t, string(huberrors.TerminalReasonContentFilter), failures.calls[0].Reason)
+	assert.Equal(t, "sentiment", failures.calls[0].Enrichment)
+	assert.Equal(t, 1, failures.calls[0].Attempts, "gave up on the first attempt, not the last")
+	assert.Equal(t, 1, metrics.outcomes["failed_final"])
+}
```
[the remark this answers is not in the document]
xernobyl: Agreed, fixed in 7c1cfcf — the fixture carries a tenant now and both fields are asserted.

This is the assertion I most wanted to have missed: `TenantID` has exactly one assignment site (`record.TenantID` in `markFailed`) and it drives both the foreign key and the tenant write lock, so it is the line a security pass leans on. Nothing bound it — the integration tests insert markers directly rather than through the worker.

xernobyl:
One thing worth settling up front, @Dhruwang, since it shapes the consumer more than it shapes this PR — `disabled_reason` is **optional and absent**, not present-and-null. Keeping it that way.

The generated type will be:

```ts
disabled_reason?: "not_configured" | "switched_off" | "no_target_language"
```

Two states rather than three, and enum-typed, so you can switch on it exhaustively for the copy. `if (status.disabled_reason)` is enough.

Two guarantees you can lean on rather than defend against:

- it is present **exactly** when `enabled` is false. Both come from one decision in `enrichment_status_service.go`, and a unit test asserts they can never disagree, so you do not need to handle "disabled with no reason" or "enabled with a reason".
- a disabled enrichment reports zeroes for `eligible`, `done`, `failed` and `failed_terminal` — the API never shows a backlog for a pipeline that will not run.

Heads up that the webhook schemas in this same spec have a `disabled_reason` that is declared `type: [string, "null"]` while the Go side omits it instead of emitting null, so a generator makes that one `string | null` *and* optional. Different shape, same name — that one is the odd one out, not this. Worth a separate ticket to fix the spec, not something to copy here.

internal/repository/enrichment_status_repository.go
```diff
@@ -78,11 +89,36 @@ const (
...
-// enrichmentCountSelect is the shared six-column SELECT list — {sentiment, emotions, translation} ×
-// {eligible, done} — used by both the per-tenant and aggregate queries so the predicates can't
+// enrichmentFailedFor builds the two failure predicates for one enrichment from its already-joined
+// marker row and its own done-predicate.
+//
+// A failure counts only while the record is still UN-ENRICHED. That is what makes the marker rows
+// advisory rather than authoritative: a later success needs no cleanup write, because a stale row
+// stops counting the moment the record is done. It also means the count can never contradict the
+// record itself.
+func enrichmentFailedFor(alias, notDone string) (failed, terminal string) {
```
Dhruwang, line 108: Failure markers are keyed only on `(feedback_record_id, enrichment)` — no target-language, model or text dimension — and are never cleared on success, while `failed` is "a marker exists AND the record is not done". Both un-enrichment paths are reachable: a `target_language` change shifts `enrichmentEffectiveTarget` so an already-translated record's `translation_lang_key` stops matching, and a `value_text` edit clears `sentiment_classified_at`/`emotions_classified_at`/`translation_lang_key`. Either resurrects a stale marker, so the endpoint reports `failed` for work that is simply re-queued — the same wrong-progress symptom ENG-2375 removes.

Cheapest fix: add `failed_at > feedback_records.updated_at` to the failure predicates. Proper fix: scope the marker to what was attempted, or clear it on a successful write.

`TestCountEnrichmentStatusCountsFailures` covers fail→succeed, which is the case that already works; the missing case is fail→succeed→invalidate.

internal/repository/enrichment_status_repository.go
```diff
@@ -366,3 +473,85 @@ func scanEnrichmentCounts(row pgx.Row, what string) (EnrichmentStatusCounts, err
...
+// countFailedRecordsAggregateSQL counts failure markers whose record is STILL un-enriched, across
+// all tenants, grouped by enrichment and permanence.
+//
+// The un-enriched condition is what keeps the markers advisory: a record that later succeeded
+// keeps its row, and that row must stop counting.
+//
+// Cheap relative to the eligible/done aggregate above. That one must scan every feedback record;
+// here the driving table is the small one — the markers, of which there are by definition few —
+// so it is a scan of failures with a primary-key lookup back to the record.
+const countFailedRecordsAggregateSQL = `
```
Dhruwang: `countFailedRecordsAggregateSQL` applies only the un-enriched test — no `tenant_settings` join for `enrichmentSentimentOn`/`enrichmentEmotionsOn`, and no `field_type`/`value_text` eligibility predicate. So `hub_enrichment_failed_records` keeps reporting failures for enrichments a tenant switched off and for records that are no longer eligible, while the endpoint reports zero for exactly those, and the neighbouring backlog gauge *is* gated.

The doc comment below documents a deliberate divergence, but a narrower one (translation's effective-target comparison); the settings and eligibility gates aren't covered by that rationale. The marker table is the driving side, so the extra join stays cheap.

internal/repository/enrichment_status_repository.go
```diff
@@ -366,3 +473,85 @@ func scanEnrichmentCounts(row pgx.Row, what string) (EnrichmentStatusCounts, err
...
+// CountFailedRecordsAggregate returns the cross-tenant failed-record counts per enrichment.
+//
+// Translation's un-enriched test is deliberately weaker than the per-tenant endpoint's: that one
+// compares against the tenant's EFFECTIVE target, which needs the settings join and the deployment
+// default. A gauge summed across tenants has no single target to compare against, so it asks only
+// whether the record has any translation at all. The consequence is that a record translated into
+// a since-changed target counts as done here while the endpoint counts it as pending — acceptable
+// for a deployment-wide gauge, and written down so the two are not mistaken for a bug when they
+// disagree.
+func (r *EnrichmentStatusRepository) CountFailedRecordsAggregate(
```
Dhruwang, line 558: `CountFailedRecordsAggregate` has three references in the tree: this definition, its doc comment, and the call site in `cmd/api/app.go`. No test at any layer — and this is the one query in the diff with no tenant predicate at all.

An integration test in `tests/enrichment_failures_test.go` seeding a switched-off tenant and a blanked-text record alongside real failures would fail on the gating issue above, so it is worth writing first.

cmd/api/app.go
```diff
@@ -788,6 +811,15 @@ func runEnrichmentBacklogPoller(
...
 		}
+
+		// Refreshed by the SAME leader on the SAME tick, so the failure gauge cannot drift from the
+		// backlog gauge and no second election is needed. Its query is cheap next to the backlog
+		// scan: it reads the markers, of which there are few, not every feedback record.
+		//
+		// queryCtx, not ctx — this shares the tick's timeout budget. Handing it the process-lifetime
+		// context would let a slow count pin a pool connection until shutdown, which is the exact
+		// thing enrichmentBacklogQueryTimeout exists to prevent.
+		refreshFailedRecords(queryCtx, statusRepo, failures)
```
Dhruwang: `refreshFailedRecords` reuses `queryCtx` — the same `context.WithTimeout(ctx, enrichmentBacklogQueryTimeout)` the backlog scan just ran under. That scan is the documented whole-table sequential scan, bounded by nothing but this context, so whatever it consumes is deducted here; on deadline error `refreshFailedRecords` calls `ClearFailedRecords()` and warns, withdrawing the gauge every tick for as long as the scan stays slow.

A fresh `context.WithTimeout(ctx, …)` for the marker count decouples them — the pool-pinning concern in the comment is satisfied by any bound, not specifically by the leftover budget.

internal/googleai/client_test.go
```diff
@@ -320,3 +320,121 @@ func TestCreateEmbedding_RateLimitReturnsRateLimitError(t *testing.T) {
...
+
+	require.NotPanics(t, func() {
+		_, terminal := terminalEmptyReason(resp)
+		require.False(t, terminal)
+	})
+
+	require.NotPanics(t, func() {
+		_ = emptyResponseDetail(resp)
+	})
+}
```
Dhruwang, line 441: The Google half of the cost-metrics feature has no behavioural test. The new tests here cover the nil-candidate and truncation fixes; none constructs a client with `WithUsageRecorder` and asserts recorded values, whereas `internal/openai/client_test.go`'s `TestUsageIsRecordedForChatCompletions` asserts Operation, Provider, Model, InputTokens, OutputTokens, ErrorType and a positive Duration. `client_factory_usage_parity_test.go` is a static AST check that a recorder is *passed*, not that anything is *recorded*.

Worth covering the gemini-vs-vertex provider value and the `ThoughtsTokenCount` contribution to output tokens, since neither has an OpenAI counterpart to piggyback on.

Dhruwang (commented):
Reviewed at `7c1cfcf`. The perf work, the tenant-boundary reasoning and the privacy posture all hold up under direct verification — the marker joins do keep `fr.tenant_id` as the boundary, the failure joins really are off the cross-tenant aggregate, and the three prior CodeRabbit threads are genuinely fixed at head, not just marked fixed.

Five things below. The one I would not merge without is the stale-marker case: it reproduces the wrong-progress symptom this PR exists to remove.

Also noted, not commented inline: terminal-failure worker behaviour is pinned for sentiment only (translation/emotions share the `markFailed` path and neither fixture sets `TenantID`); `recordUsage`/`errorTypeOf` are ~90% duplicated between the two clients while `internal/llm/usage.go` is types-only; `markFailed`'s default error branch logs without incrementing any counter; and `emptyResponseDetail` interpolates `BlockReasonMessage` unbounded where the OpenAI refusal is bounded at `%.256s` — pre-existing, but migration 022's comment claims provider text is "already truncated at the client", which is true of one path and not the other.

internal/repository/enrichment_status_repository.go
xernobyl: You are right, and thank you for gating on it — fixed in a48e03e, though not the way either of us first suggested.

I tried the cheap fix and it is the wrong shape. `updated_at` is RECORD-level while markers are per `(record, enrichment)`, so a successful sentiment write bumps the timestamp for the emotions and translation markers too and silently drops them from the counts. The smoke run caught it immediately: **7% of records fell out of the accounted set** — the failure this whole thing exists to prevent, arriving through the fix for it. Any per-enrichment currency rule would need a per-enrichment timestamp, and only emotions has one.

So it is your second option: the successful write deletes the marker it resolved, in the transaction it already opens, against the primary key, a no-op probe when there is nothing there. That gives up the write-free-on-success property the design was built around — correctly, since that property rested on the same bad assumption you spotted.

The integration test walks fail → succeed → invalidate, and the trap is written down next to the predicate so the next person reaching for the timestamp finds the measurement first.

internal/repository/enrichment_status_repository.go
xernobyl: Agreed — fixed in a48e03e. Eligibility and the tenant switch are now applied, so it matches the endpoint and the neighbouring backlog gauge.

I left translation's weaker un-enriched test as-is and kept the rationale, but tightened the comment to say plainly that the excuse covers only that one divergence and not the gates — which was the actual gap you were pointing at.

internal/repository/enrichment_status_repository.go
xernobyl: Fair, and the ordering advice was right — writing the test first is what surfaced the gating bug as a concrete number rather than a code reading.

`TestCountFailedRecordsAggregateIsGated` in a48e03e seeds a switched-off tenant, a blanked-text record and a live failure, and also asserts the gauge agrees with what the endpoint reports for the same tenants. Each gate was verified by reverting it on its own.

cmd/api/app.go
xernobyl: Fixed in a48e03e — it gets its own `context.WithTimeout(ctx, enrichmentBacklogQueryTimeout)`.

You are right about the reasoning too: the comment defended the bound, and the bound was never the part in question. I had swapped an unbounded context for a coupled one in an earlier pass and did not notice I had traded one problem for another.

internal/googleai/client_test.go
xernobyl: Fixed in a48e03e. `TestUsageIsRecordedForGenerateContent` drives the real genai SDK against a stub and asserts operation, provider, model, both token counts, the bounded `error.type` and a positive duration — the same shape as the OpenAI test.

Both of the cases with no OpenAI counterpart are covered: `ThoughtsTokenCount` folding into output (4 completion + 96 thinking = 100, since a Pro model falling back to its default thinking budget is where the bill actually lands), and the gemini-vs-vertex provider values, which matter because the two are billed separately. Embeddings assert duration with zero tokens, so the empty token series reads as the provider's doing rather than a wiring bug.

xernobyl:
All five addressed in `a48e03e` — CI green, and the end-to-end run is back to 916 checks / 0 failures.

One is worth calling out because it did not go the way the comment suggested. I took the cheap fix first (`failed_at > updated_at`) and it is actively wrong: `updated_at` is record-level while markers are per `(record, enrichment)`, so a successful sentiment write invalidates the emotions and translation markers too. The smoke run put a number on it — **7% of records dropped out of the accounted set**, which is precisely the symptom the ticket exists to remove. So it is the second option instead: the successful write deletes the marker it resolved. That trades away the write-free-on-success property the design was built around, which is the right trade, since that property rested on exactly the assumption you found to be false.

Also folded in the four items from your review body: `BlockReasonMessage` is truncated to match the OpenAI refusal cap (and migration 022's comment no longer claims something true of only one path), the marker-write failure path increments a counter instead of only logging, the translation and emotions fixtures carry a tenant with each pipeline pinning its own marker wiring, and the duplicated recorder shape moved into `internal/llm`.

Back to you when you have a moment.

Dhruwang (approved):

PR #127, opened 2026-08-20
feat(sdk): generate and publish @formbricks/hub from this repo (ENG-2092)

xernobyl:
## What does this PR do?

Replaces the Stainless-generated `@formbricks/hub` SDK with one generated here from `openapi.yaml` by [`@hey-api/openapi-ts`](https://heyapi.dev/), published to npm on a stable Hub release.

Linear: https://linear.app/formbricks/issue/ENG-2092 · Decision doc: [Hub SDK after Stainless — why Hey API](https://linear.app/formbricks/document/hub-sdk-after-stainless-the-decision-and-why-hey-api-c5160458e30e)

**Why.** Stainless shuts down 1 September 2026. Of six generators tested against this spec, Hey API was the only free one that cleared every requirement — and Stainless was the only one of the six that got our repeated array filters *wrong*, which is why `formbricks/formbricks` carries ~440 lines correcting it.

**The SDK is a build artifact.** The generator config is committed; the output is not. `src/generated/` and `dist/` are gitignored, produced in CI, and published on release. That is defensible here for three specific reasons, and stops being defensible if any one stops holding:

1. the generator is deterministic — verified byte-identical across runs, which is what the skip-if-unchanged check depends on;
2. npm provenance names the exact source commit, so anyone can regenerate and compare;
3. the tarball ships readable `src/` (verified: 30 files, 144 KB packed).

**Publishing is credential-free.** npm trusted publishing (OIDC) — no token, and unlike the script this replaces, no token *fallback*. Provenance is automatic for a public package from a public repo.

**Two jobs, on purpose.** The build job runs the generator and its transitive dependencies and holds **no** `id-token`; a separate publish job downloads the publishable package, checks it, and runs `npm publish --ignore-scripts`. Without that split, one compromised dev dependency could publish arbitrary code as `@formbricks/hub` — to ~10k downloads/month, carrying genuine provenance, which makes it harder to spot rather than easier. What the publish job checks before it spends the credential is described under *Review round* below.

**Version starts at 0.14.0** and the SDK version line stays independent of the Hub's (they have never matched — npm was at 0.13.0 while the Hub was at 0.9.x). npm caret ranges pin the minor on `0.x`, so `^0.13.0` consumers stay on 0.13.x and are *not* dragged into the method renames.

Also in here: `stainless-action.yml` deleted, and the three `x-stainless-model` extensions removed from `openapi.yaml` — verified they were doing nothing for us (see the commit for what was checked).

### The rename, for consumers

Resource methods become one flat function per operation, so bundlers can tree-shake:

```diff
- await client.feedbackRecords.list(params)
+ await listFeedbackRecords({ client, query: params })
```

`createHubClient({ apiKey, baseUrl })` is provided as sugar; `baseUrl` is required rather than defaulted because the Hub is self-hostable. Zod validators for every shape live behind `@formbricks/hub/schemas` so the default entry pulls in **zero** dependencies and `zod` stays an optional peer.

The full 15-method → 32-function mapping is in the job summary of the `sdk-preview` workflow, and is what ENG-2351 will migrate against.

## Review round (2026-09-15)

Bhagya found five issues, all reproduced before fixing. Summarised here because they changed the shape of the pipeline, not just its details:

- **The credential-bearing job could execute artifact-controlled code.** `npm publish` runs lifecycle hooks out of the manifest it is handed, and an injected `prepublishOnly` was confirmed to execute in the job holding the OIDC token. Now `--ignore-scripts`, plus a validation step that rejects the manifest on unexpected name, version, any publish lifecycle hook, or any `publishConfig` key other than `access`.
- **The release compared only `src/`**, so a change confined to the exports map, peer ranges, engines, bundler output or the README was invisible to the publish decision. It now compares the **packed** package, manifest field by field, with `version` the one deliberate exception.
- **The preview's surface diff grepped the source tree**, missing `export class`, default exports, `export { }` and `export *`, and ignoring the export map. It now reads the export map and the built entry points through the TypeScript compiler API — and compares **per export condition**, since unioning them let a name dropped from `import` but kept in `require` report as no change.
- **Concurrency was keyed per release**, so two releases could compare against the same npm `latest` and race. One package-wide group now.
- **The provenance check tested an exit status** that is `0` with empty output when the field is absent. It parses the attestation and fails if it is not there.

Three things not asked for, added because fixing the above exposed them: the publish job **re-packs the artifact and requires the integrity hash the build job recorded**, so what reaches npm is what was compared and tested; the packing and npm error handling is **shared between both workflows** rather than duplicated; and both comparison helpers now have **tests, each confirmed to fail when its own fix is reverted**.

## Prerequisite: npm trusted publishing must be repointed (needs @mattinannt)

`@formbricks/hub@0.12.0` was published via OIDC bound to `formbricks/hub-typescript` + `publish-npm.yml`. npm allows **one** trusted publisher per package, matched on repository *and* workflow filename, case-sensitive. Until it is repointed, the publish job fails with "Unable to authenticate".

On npmjs.com → `@formbricks/hub` → Settings → Trusted Publisher:

| Field | Value |
|---|---|
| Repository | `formbricks/hub` |
| Workflow filename | `publish-sdk.yml` |
| Environment | `npm-publish` |
| Allowed actions | `npm publish` |

- **Environment is not optional in practice.** The binding does not include a branch or tag, so without it anyone with push access could publish arbitrary content from any branch with genuine provenance. A GitHub Environment named `npm-publish` with a deployment-branch rule (tags / default branch) is what closes that. The workflow already references it.
- **Allowed actions must be set explicitly** — configurations created after 2026-05-20 require it, and ours is new.
- **Repointing removes `hub-typescript`'s ability to publish** `@formbricks/hub`, so do it at cutover. `@formbricks/hub-mcp` is unaffected (separate config, token-published).
- **Afterwards**, set "Require two-factor authentication and disallow tokens" on the package. It does not affect trusted publishers and closes the residual old-token path.

Maintainers are `formbricks-com` and `matthiasnannt`; I don't have access.

**Order, and one thing worth knowing.** Merging publishes nothing: `publish-sdk.yml` triggers only on `release: [published]` and manual dispatch, and nothing in this repo auto-creates a release. So the sequence is: create the `npm-publish` environment → repoint npm → merge → **run the workflow manually with `dry_run: true`** → cut a release.

That dry run matters because `workflow_dispatch` only reaches workflows that exist on the default branch, so this one **cannot be exercised at all until it merges** — the preview workflow is the only half that has run in CI. A dry run does the full build, compare and artifact upload while skipping the publish job entirely, and holds no credentials.

The two prerequisites also fail in opposite ways: a wrong npm repoint fails **loudly** at the publish step with nothing published, but a missing `npm-publish` environment fails **silently** — GitHub auto-creates it without protection rules and the publish succeeds, leaving exactly the any-branch hole the environment exists to close.

## How should this be tested?

All of this runs locally from `sdks/typescript`, with no secrets and no database:

```bash
cd sdks/typescript
pnpm install --ignore-scripts
pnpm build          # generate + dual ESM/CJS build
pnpm test           # builds, then runs the wire tests
./node_modules/.bin/tsc --noEmit
```

1. **32 operations generated**, one per `operationId` in `openapi.yaml`, as flat functions with no class. `tsc --noEmit` reports 0 errors.
2. **The regression test is the point.** `tests/wire.test.mjs` drives the *built* package (so it covers the exports map) against a local echo server and asserts `{ source_type: ["survey","review"] }` goes out as `?source_type=survey&source_type=review`. I confirmed it fails on the old comma-joined output rather than passing regardless — replay `?source_type=survey%2Creview` through its assertions and all three reject it.
3. **Determinism**: `pnpm generate` twice and `diff -r src/generated` — must be byte-identical. Everything else depends on this.
4. **Tarball hygiene**: `npm pack --dry-run --json` — 30 files, nothing outside the `files` allowlist (`dist`, `src`, `README.md`, `LICENSE`), and readable `src/` present.
5. **Spec cleanup didn't regress anything**: after removing the `x-stainless-model` extensions, still 32 operations, `children?: Array<TaxonomyNodeData>` intact, `TaxonomyRunData` and `EnrichmentTypeStatus` still single shared types. `npx @stoplight/spectral-cli@6 lint openapi.yaml` → no errors.
6. **The pipeline, without publishing**: run `publish-sdk.yml` via workflow_dispatch with `dry_run: true`. The build job should run end to end and the publish job should be skipped.
7. **The skip path**: with the spec unchanged from 0.12.0's, the compare step should report "identical — skipping publish" rather than publishing.
8. **The guard path**: set `version` back to `0.12.0` with a changed surface — the build must fail with "version 0.12.0 is already on npm. Bump the version".
9. **`sdk-preview` on this PR** should print the surface diff to its job summary and request no write permissions.

Not testable before the npm change: the real publish. Treat the 0.9.0 Hub release as the cutover and watch the job.

### Required

- [ ] Ran `make build` — no Go code changed
- [ ] Ran `make tests` — not applicable; the SDK has its own suite (`pnpm test`), which passes 3/3
- [x] Ran `make fmt` and `make lint`; no new warnings — no Go changes; Prettier run over the new TS/JSON/MD
- [x] Merged the latest changes from main onto my branch — cut from `origin/main`, 0 behind
- [ ] If database schema changed — no schema change

### Appreciated

- [x] If API changed: added or updated OpenAPI spec and ran contract tests — `openapi.yaml` changed only by removing vendor extensions; Spectral clean, no behaviour change
- [ ] If API behavior changed — none
- [x] Updated docs in `docs/` if changes were necessary — `AGENTS.md` gains a TypeScript SDK section; the npm-facing README is `sdks/typescript/README.md`
- [ ] Ran `make tests-coverage` — no Go logic changed

---

**Note on conflicts:** #126 (ENG-2369, docs off Stainless) also edits `AGENTS.md`. I placed this section higher in the file to avoid a conflict, but whichever lands second may still need a trivial rebase.


xernobyl:
Marked ready so CodeRabbit actually reviews it — it skips drafts, so its green tick until now meant "didn't look", not "looks fine".

**Please don't merge this before the npm trusted-publisher change lands.** Nothing here breaks on merge, but the first Hub release after it would try to publish and fail authentication, which is a confusing way to find out. The four values needed are in the description; @mattinannt is doing it since maintainers are `formbricks-com` and `matthiasnannt`.

Ordering that works: npm repointed → this merges → next Hub release publishes `@formbricks/hub@0.13.0`.

xernobyl:
@coderabbitai review

xernobyl:
Brought this up to date after three weeks. Conflict resolved, toolchain current, and **one thing changed that is worth reading rather than skimming**.

**The version had to move: 0.13.0 → 0.14.0.** Stainless published one final release under `0.13.0` from `formbricks/hub-typescript` on 26 Aug, just before the platform shut down — provenance confirms the repo and tag, and the tarball carries the old resource-method export shape (`./resources/*`, `api-promise`). So the number this branch reserved now means *the old SDK* to anyone installing it, and this PR's own publish guard would have refused it — which is the guard working as intended rather than a problem.

0.14.0 keeps the original reasoning intact: npm caret ranges pin the minor on `0.x`, so `^0.13.0` consumers stay on the Stainless shape until they deliberately opt in.

**Conflict.** Only one — `stainless-action.yml`, deleted here and modified on `main`. Kept the deletion: main's change was the ENG-3213 sweep mechanically bumping the action SHA and a version comment, not a decision to keep it, and the platform it talks to has been down since 1 Sep. `AGENTS.md` and `openapi.yaml` merged clean — the docs section from #126 and the SDK section here sit apart in the file, which is why the earlier re-anchoring was worth doing.

**The SDK is now 32 operations, not 31** — `main` added `retry-tenant-enrichments` (#125/#128 line of work) and the generator picked it up with no changes needed.

**Toolchain**, matched to what ENG-3213 chose for the repo: pnpm 11.5.0 → 12.4.1, tsdown 0.22.14 → 0.23.0, zod → 4.6.4, TypeScript 5.9.3 → 6.0.3. `@hey-api/openapi-ts` and prettier were already current.

Two findings from doing it:

- **TypeScript stops at 6, for a second and independent reason.** The known blocker is `@astrojs/check` peering `^5 || ^6` for the docs site, which does not apply to this package. But 7.0.2 fails here too: generation dies with `Cannot read properties of undefined (reading 'AnyKeyword')` because the native port changed the compiler-API export shape and the generator reads `ts.SyntaxKind` off the default import. So the ceiling is any tool embedding the compiler API, not just Volar-based ones.
- **zod is a range, not a pin, deliberately.** Pinning `4.6.5` made pnpm 12 write a `minimumReleaseAgeExclude: [zod@4.6.5]` bypass into a new `pnpm-workspace.yaml` instead of failing — 4.6.5 was 15 hours old, inside its own cooldown. A range lets the cooldown work: pnpm resolved 4.6.4 and wrote no bypass. Worth keeping for a dependency of a package we publish.

Re-verified end to end on the merged spec: 32/32 operations, `tsc --strict` clean, byte-identical across two generate runs (the skip-if-unchanged gate depends on that), wire tests 3/3 including the array-filter regression, every `exports` target present on disk, and a 30-file / 154 KB tarball with nothing outside the `files` allowlist. `go build` and Spectral both still clean after the merge.

**The npm prerequisite is unchanged and still the only blocker to releasing** — the trusted publisher needs repointing to `formbricks/hub` + `publish-sdk.yml` + the `npm-publish` environment + `npm publish`. Also worth checking whether `hub-typescript` is still the configured publisher, since it evidently was on 26 Aug.

xernobyl:
@BhagyaAmarasinghe one infra item that I can't do myself, and it's the part with a sharp edge.

The publish job runs in a GitHub Environment called `npm-publish`. `formbricks/hub` currently has **zero** environments, and creating one needs admin — I only have push. So it's yours or Matti's.

Why it needs to exist *before* the first release rather than after: npm's trusted-publisher binding covers repo + workflow filename but **not the branch or tag**. The environment is what adds that constraint. And referencing an environment that doesn't exist auto-creates it with no protection rules — so the first release would publish fine and quietly leave the any-branch hole open. Nothing would look broken, which is the annoying bit.

What it needs: **Settings → Environments → `npm-publish`**, plus a deployment branch/tag rule limiting it to tags and the default branch. No secrets or variables — the publish authenticates over OIDC, there's no `NPM_TOKEN` anywhere and there shouldn't be one.

Ordering overall is: environment → Matti repoints the npm trusted publisher → merge. Happy to be told the branch rule should be shaped differently, that's more your call than mine.

xernobyl:
@coderabbitai review

.github/workflows/publish-sdk.yml
```diff
@@ -0,0 +1,225 @@
...
+      # --provenance is passed explicitly even though trusted publishing is
+      # documented as generating attestations automatically: the two npm docs
+      # pages disagree on whether the flag is needed, and passing it when it is
+      # already implied is harmless while omitting it when it is not is silent.
+      - name: Publish
+        working-directory: package
+        run: |
+          set -euo pipefail
+          echo "Publishing $(node -p "require('./package.json').name")@$(node -p "require('./package.json').version")"
+          npm publish --access public --tag latest --provenance
```
BhagyaAmarasinghe: Could we publish with `--ignore-scripts` and validate the downloaded manifest before this step? The build job uploads `package.json` after running generator/build dependencies, and `npm publish` executes lifecycle hooks from that artifact by default. I reproduced an injected `prepublishOnly` hook executing with this command. Because this job has `id-token: write`, artifact-controlled code regains execution in the credential-bearing job, defeating the isolation this split is meant to provide. Please also require the expected package name/version and reject lifecycle scripts.

.github/workflows/publish-sdk.yml
```diff
@@ -0,0 +1,225 @@
...
+            exit 0
+          fi
+
+          PUBLISHED="$(npm view @formbricks/hub version)"
+          echo "Published latest: $PUBLISHED / building: $VERSION"
+
+          npm pack "@formbricks/hub@$PUBLISHED" --pack-destination . >/dev/null
+          tar -xzf ./*.tgz            # extraction only: never install or execute it
+
+          if diff -r package/src ../src >/dev/null 2>&1; then
```
BhagyaAmarasinghe: This unchanged check only considers `src`, while the published artifact also includes `dist`, `package.json`, `README.md`, and `LICENSE`. A version bump that changes the exports map, peer dependencies, engines, bundler output, or package documentation with identical source will return here and never publish. I reproduced different `package.json` hashes while this source-tree comparison still exited 0. Please compare normalized packed contents, excluding only fields that are intentionally expected to differ.

.github/workflows/sdk-preview.yml
```diff
@@ -0,0 +1,114 @@
...
+            echo "Package not published yet; nothing to compare." >> "$GITHUB_STEP_SUMMARY"
+            exit 0
+          fi
+
+          PUBLISHED="$(npm view @formbricks/hub version)"
+          npm pack "@formbricks/hub@$PUBLISHED" --pack-destination . >/dev/null
+          tar -xzf ./*.tgz          # extraction only: never installed or executed
+
+          exports_of() {
+            grep -rhoE '^export (const|function|type|interface) [A-Za-z0-9_]+' "$1" 2>/dev/null \
```
BhagyaAmarasinghe: This does not measure the package's public surface: it scans every source file, counts internal declarations, and misses `export class`, default exports, `export { ... }`, and `export *`. It also ignores `package.json.exports`. Against the current 0.13.0 package, this misses the public `FormbricksHub` and default exports entirely, while the export map changes from 31 entries to 2. Please derive the comparison from the built entry-point declarations/runtime exports plus the package export map.

.github/workflows/publish-sdk.yml
```diff
@@ -0,0 +1,225 @@
...
+        description: "Build and compare, but do not publish"
+        type: boolean
+        default: true
+
+# Least privilege by default; the publish job opts into id-token explicitly.
+permissions:
+  contents: read
+
+concurrency:
+  group: publish-sdk-${{ github.event.release.tag_name || github.ref }}
```
BhagyaAmarasinghe: Including the release tag/ref gives different releases different concurrency groups, so two stable releases can build concurrently, both compare against the same npm latest version, and race to publish the same target version. Please use one package-wide concurrency group so compare and publish are serialized across all SDK releases.

.github/workflows/publish-sdk.yml
```diff
@@ -0,0 +1,225 @@
...
+          set -euo pipefail
+          echo "Publishing $(node -p "require('./package.json').name")@$(node -p "require('./package.json').version")"
+          npm publish --access public --tag latest --provenance
+
+      - name: Confirm provenance landed
+        run: |
+          set -euo pipefail
+          VERSION="${{ needs.build.outputs.version }}"
+          for i in $(seq 1 10); do
+            if npm view "@formbricks/hub@$VERSION" dist.attestations >/dev/null 2>&1; then
```
BhagyaAmarasinghe: `npm view` exits 0 and prints nothing when an existing package has no such field, so this condition succeeds on the first iteration even when no attestation exists. I confirmed that behavior against a package without `dist.attestations`. Please capture the output and require a non-empty, valid provenance object before exiting successfully.

BhagyaAmarasinghe (changes requested):
Requesting changes because the credential-bearing publish job can execute artifact-controlled lifecycle scripts, and the release and preview comparisons do not cover the actual published package/public API. The SDK generation and runtime checks otherwise passed at this exact head.

.github/workflows/publish-sdk.yml
xernobyl: Fixed — and I reproduced it first to be sure I was fixing the right thing. With an injected `prepublishOnly` in the manifest, `npm publish --dry-run` ran the hook; adding `--ignore-scripts` stopped it.

Both halves are in now. The publish runs `--ignore-scripts`, and a validation step ahead of it rejects the manifest outright: expected name, expected version (from the build job's output, so the two have to agree), and no publish lifecycle hooks — `prepublish`, `prepare`, `prepublishOnly`, `prepack`, `postpack`, `publish`, `postpublish`.

On `publishConfig` I went a bit wider than you asked. Rather than checking `registry` alone, it now rejects any key except `access`: `registry`, `tag` and `provenance` can all steer a publish from there, and closing the field seemed better than closing the three keys I happened to think of. The registry is named on the command line too.

One thing your comment made me notice: nothing actually tied the artifact to what the build job compared and tested. The publish job now re-packs the download and requires the integrity hash the build job recorded. `npm pack` turns out to be reproducible for a given tree and npm version — I checked, and a re-pack of the extracted tarball matches — so both jobs now pin the same npm. Tampering with `dist`, adding a `bin` field, and dropping the README each fail it; I tried all three.

Worth being explicit about the limit, and it's in the comment too: that's an integrity check, not a trust check. A compromise *inside* the build job would produce a matching hash, because it produces both. Keeping its code out of this job is what `--ignore-scripts` and the validation are for.

.github/workflows/publish-sdk.yml
xernobyl: Fixed — it compares the packed contents of both sides now, not `src`.

`scripts/compare-packed.mjs` diffs every shipped path byte for byte and the manifest field by field, printing `A`/`D`/`M` per path and naming which manifest fields differ. Against 0.13.0 it reports `M package.json (fields: devDependencies, engines, exports, files, main, module, peerDependencies, …)` — the class you reproduced.

`version` is the one deliberate exception: bumping it is the input to a release, not evidence the package changed, so normalising it out is what lets an otherwise unchanged package still skip. That cuts both ways and is easy to get wrong later, so both directions are pinned by tests.

While I was in there I checked the thing that would have quietly broken the skip path anyway: whether npm injects fields into a published manifest (`gitHead` and friends). It doesn't on current npm — I packed 0.13.0 and diffed the key set — so identical can genuinely compare identical rather than always differing by one field nobody put there.

.github/workflows/sdk-preview.yml
xernobyl: Fixed. `scripts/surface.mjs` reads `package.json.exports` subpath by subpath, and for each entry point pulls the exported names out of the built declaration and runtime files with the TypeScript compiler's `getExportsOfModule` — so `export class`, default exports, `export { ... }`, `export *` and type-only exports all count, and internal declarations the entry point doesn't re-export don't.

Against 0.13.0 it now reports exactly what you said was missing: `FormbricksHub` and `default` under removed runtime *and* type names, `Uploadable` as type-only, and the export map going 31 subpaths → 2 (`./schemas` added, `.` changed).

Two things I changed beyond what you flagged, both from re-reading the output:

- the removed list is the breaking one, and it was the list being truncated at 30 with no way to expand — while the harmless "added" side got a full expandable dump. Backwards. Both expand now.
- a broken exports map fails the job for our generated package, but only leaves a note for the package on npm. Nothing in a PR can fix something already published, and I didn't want a bad upstream release turning every preview red for a reason nobody could action.

.github/workflows/publish-sdk.yml
xernobyl: Fixed — one package-wide group, `publish-sdk`, with `cancel-in-progress` still `false` so a queued release waits rather than being dropped.

.github/workflows/publish-sdk.yml
xernobyl: Fixed, and confirmed against the registry rather than taking it on faith: `npm view is-odd@3.0.1 dist.attestations --json` exits 0 and prints nothing, exactly as you described.

It captures the output now and requires a parseable object with a `url` under `registry.npmjs.org/-/npm/v1/attestations/` and a `provenance.predicateType` of `https://slsa.dev/provenance/v1` — the shape 0.13.0 actually returns, which is what I checked it against. And it fails rather than warns when that doesn't show up within 100s, since provenance is most of the reason this publishes from CI instead of a laptop.

xernobyl:
Thanks @BhagyaAmarasinghe — all five were real, and each one I reproduced before fixing it so I knew I was fixing the right thing. Replies are on the individual threads; the short version, plus what re-reviewing my own work turned up.

**Your five.** `npm publish --ignore-scripts` plus a validation step that rejects the manifest by name, version and publish lifecycle hooks (I confirmed an injected `prepublishOnly` executes without it, and doesn't with it). The release now compares **packed contents** rather than `src`, manifest field by field. The preview derives its surface from the **export map and the built entry points** through the TypeScript compiler API, so `export class`, `default`, `export { }` and `export *` all count — against 0.13.0 it now shows `FormbricksHub`, `default` and the 31→2 export-map change. One package-wide concurrency group. And the provenance check parses `dist.attestations` instead of trusting an exit status that is 0 with empty output.

**Three things I changed that you didn't ask for**, because fixing yours exposed them:

1. **Nothing tied the published bytes to the compared bytes.** A comment claimed the packed tarball was what the publish job sends; nothing enforced it. `npm pack` turns out to be reproducible for a given tree and npm version, so the publish job now re-packs the artifact and requires the build job's integrity hash. Tampered `dist`, an injected `bin`, a dropped README — all fail it. It's an integrity check, not a trust check, and the comment says so: a compromise *inside* the build job produces both sides.
2. **The fixes duplicated ~25 lines of subtle npm handling across both workflows.** That's now `prepare-comparison.mjs`, shared — so the release decision and the preview you read measure the same two packages the same way, rather than drifting apart the first time one gets patched.
3. **The breaking half of the surface report was the truncated half.** Removals cut off at 30 with no way to expand while additions got a full dump. Both expand now. And a broken exports map fails for our package but only notes for the one on npm, since no PR can fix something already published.

**Tests.** The failure mode here is a check that quietly stops checking, so `compare-packed` and `surface` now have coverage — and I verified each test **fails when its own fix is reverted**, rather than just passing. 14/14 green.

**One I deliberately didn't do:** both jobs run harden-runner with `egress-policy: audit`, not `block`. `block` is the stronger posture for a job holding an OIDC token, but the allowlist has to cover the registry, Sigstore and GitHub's OIDC/artifact APIs, and getting it wrong breaks the one path we can't rehearse before the npm repoint. Parked as its own change, best done once a first successful publish gives us a real egress log. Happy to be overruled if you'd rather it went in here.

Still not merge-ready for the two reasons in the description — the `npm-publish` environment doesn't exist yet (needs admin) and npm still points at `hub-typescript`.

xernobyl:
One known-unknown on the integrity gate, flagged rather than buried, since you'd be the one debugging it.

The CI preview packs `0.14.0` to `sha512-RuFFbw…`; my laptop packs the same commit to `sha512-3FNsy2…`. That difference is fine and expected — different npm version and platform — and the gate only ever compares CI to CI, with both jobs pinned to the same npm. But it does mean the property the gate relies on is "`npm pack` is stable for a fixed tree *and* a fixed npm", which I verified locally (two packs, plus a re-pack of the extracted tarball, all identical) and cannot fully verify across `upload-artifact` → `download-artifact` until a real release runs.

The specific thing that could bite: if the artifact round-trip alters file metadata that ends up in the tarball — modes being the plausible one — a legitimate release would fail the gate. I think that's unlikely (everything we ship is 0644, and the download normalises to that anyway), and the failure direction is the safe one: the release stops with expected-vs-actual printed, nothing gets published. If you see it fail on the first real release with a package that otherwise looks right, that's the cause to check first, not a tamper.

Worth knowing before the first publish rather than during it.

.github/workflows/sdk-preview.yml
```diff
@@ -0,0 +1,94 @@
...
+    name: Diff the SDK surface
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    defaults:
+      run:
+        working-directory: sdks/typescript
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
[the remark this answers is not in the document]
xernobyl: Valid — fixed on both checkouts in 2004c48.

Both jobs run the generator and its dependencies, and neither does an authenticated git operation after checkout, so there's nothing to lose by dropping the token. `docs-build.yml` already does exactly this with the same reasoning; these two were the outliers rather than the precedent.

sdks/typescript/scripts/prepare-comparison.mjs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+function extract(tarball, into) {
+  execFileSync("tar", ["-xzf", tarball, "-C", into, "--strip-components=1"], {
+    stdio: ["ignore", "inherit", "inherit"],
+  });
+}
+
+// Start from a clean tree so nothing from an earlier run is counted as content.
+rmSync(workdir, { recursive: true, force: true });
```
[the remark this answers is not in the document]
xernobyl: Fair, and fixed in 2004c48 — it now removes only the four directories it creates, rather than the caller-supplied `workdir`.

Both CI callers pass `.compare` so it wasn't reachable, but `node scripts/prepare-comparison.mjs .` would have deleted the package, and that's a sharp edge with no reason to keep.

sdks/typescript/scripts/surface.mjs
```diff
@@ -0,0 +1,339 @@
...
+
+      const symbol = checker.getSymbolAtLocation(sourceFile);
+      if (!symbol) continue;
+
+      for (const exported of checker.getExportsOfModule(symbol)) {
+        const name = exported.getName();
+        // Interop marker emitted by transpiled CommonJS, not part of the API.
+        if (name === "__esModule") continue;
+        (DECLARATION_FILE.test(relativePath) ? types : runtime).add(name);
+      }
```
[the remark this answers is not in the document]
xernobyl: Good catch, and a real false negative — fixed in 2004c48.

`entryFiles` now keys each subpath by condition path (`import.default`, `require.types`, …) and `buildSurface` reads names per condition instead of unioning them, so a name dropped from `import` while `require` keeps it is reported rather than absorbed. Conditions that agree are merged back for display, so the ordinary case is still one line per subpath and a divergence stands out precisely because it fails to merge. There's a test, and I confirmed it fails against the previous union behaviour rather than just passing against the new one.

Worth flagging what the fix exposed, since it's a limitation rather than a bug: once conditions are compared exactly, a package whose export map has been *restructured* has no counterpart conditions to compare at all. 0.13.0 is that case — its ESM entry sits at the top-level `default` while ours is under `import.default` — so those names genuinely can't be diffed. Rather than leave an empty section that reads as "nothing changed here", the report now says which conditions exist on only one side and points at the export maps. The alternative, normalising `default` onto `import.default`, guesses at resolution order, and I'd rather be honest about what wasn't compared than quietly compare the wrong pair.

xernobyl:
@coderabbitai review

mattinannt:
@coderabbitai please review

.github/workflows/publish-sdk.yml
```diff
@@ -0,0 +1,419 @@
...
+  publish:
+    name: Publish to npm
+    needs: build
+    if: needs.build.outputs.should_publish == 'true' && !(github.event_name == 'workflow_dispatch' && inputs.dry_run)
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    # Gates the OIDC token behind branch/tag protection rules. npm's
+    # trusted-publisher binding does not include a ref, so this environment is
+    # what stops any branch from publishing.
+    environment: npm-publish
```
BhagyaAmarasinghe: [P1] Please create and protect this environment before approving the release path. I rechecked the live repository at this exact head and `GET /repos/formbricks/hub/environments` still returns `total_count: 0`. The npm trusted-publisher binding does not constrain the ref, so this job relies on the environment deployment policy to prevent a non-dry run from another selectable ref from obtaining `id-token: write` and publishing with genuine provenance once npm is repointed. Referencing a missing environment can create it without the intended branch/tag restrictions. Please create `npm-publish`, restrict it to the default branch and protected release tags, and verify the npm trusted-publisher binding before approval.

BhagyaAmarasinghe (changes requested):
Requesting changes because the live repository still has no GitHub Environments, while the publish job relies on `npm-publish` to constrain which refs can obtain the npm publishing identity. The earlier code findings are addressed at this head; this external deployment guard must be created, protected, and verified before approval.

PR #128, opened 2026-08-21
feat: reconcile enrichment coverage instead of hoping events cover it (ENG-2376)

xernobyl:
## What does this PR do?

~~Stacked on #125~~ — **#125 merged**, and this branch is rebased onto `main`, so the diff is the reconciler alone. One rebase casualty worth knowing: the cooldown migration renumbered from 023 to **024**, because #130 took 023 on main while this was in flight.

ENG-2375 made enrichment coverage *reportable*: `eligible`, `done`, `failed`, `failed_terminal`, and a terminal-vs-transient taxonomy so the two kinds of failure are told apart. This makes it *self-healing*.

Enrichment coverage becomes a level-triggered invariant — **every eligible record ends up enriched, or classified as something the provider will never accept**. The event path stays exactly as it is and keeps enrichment fast; a periodic sweep finds what that path missed and is what makes it eventually complete. Events are the optimisation, reconciliation is the guarantee.

That subsumes four problems that currently look separate: records stranded because credentials were unset when they arrived (Dhru's repro on ENG-2375), transient failures that used up their retry budget, jobs lost to a crash between enqueue and work, and [ENG-2243](https://linear.app/formbricks/issue/ENG-2243)'s emotions phantom-pending records, which are exactly the pending set.

### The pieces

**The pending set** reuses the status endpoint's own predicate constants rather than restating them. That is deliberate and load-bearing: the endpoint reports `eligible - done - failed` as work that is scheduled, and this is what schedules it, so the two definitions of "pending" cannot be allowed to drift. Terminal failures are excluded — without that the sweep re-runs a content-filtered record on every tick forever, at a provider call each time, and never terminates. This is why the taxonomy in #125 had to land first.

**A backfill queue per enrichment**, separate from the live queue the event path feeds. Separation *is* the rate control: the two lanes draw from different `MaxWorkers` budgets, so a sweep of a large backlog cannot make a record submitted a moment ago wait behind it. The alternative — pacing inserts onto the live queue — means guessing a rate that matches drain speed, and guessing low is indistinguishable from the reconciler not running. It also fixes something older: `backfill-classify` enqueues onto the **live** queues today, so running one already starves live enrichment.

**Top up to a target depth, not at a rate.** Each tick adds only what the workers have already drained, so `river_job` stays bounded whether the backlog is a thousand records or fifty million.

**Uniqueness across the in-flight states**, where both halves matter in opposite directions. `completed` must stay **out** — River's default set includes it, and with no `ByPeriod` the window is unbounded, so the first sweep of a record would be the only one that ever ran. `retryable` must stay **in** — a job waiting out its backoff runs again by itself, and re-enqueueing it doubles the provider calls for a record already being handled. The tenant purge declared the same set for the same reasons, so they are one helper now rather than two lists that can drift.

Worth not conflating: the queue-depth query *excludes* `retryable`, because a job waiting out a backoff is not occupying a worker — counting it would have the sweep see a full queue while nothing drains.

**Scheduled by River's periodic scheduler**, which runs from the elected leader only. That is the reason for the choice: with several hub-workers deployed, a ticker in each process would have every replica sweeping at once. The scheduler's state is in memory, so a restart or leader change restarts the cycle — which costs nothing, because the sweep is level-triggered: a missed tick is invisible and an extra one finds the queue at depth and does nothing.

### Config

`ENRICHMENT_RECONCILE_ENABLED` (kill switch, on by default), `_INTERVAL_SECONDS` (300), `_TARGET_DEPTH` (1000), and `{TRANSLATION,SENTIMENT,EMOTIONS}_RECONCILE_MAX_CONCURRENT` (2). Interval and depth fall back to their defaults when misconfigured, because a zero interval makes River reject the job outright and a zero depth makes the sweep enqueue nothing — both take the reconciler silently offline, which is the failure this feature exists to prevent.

### The retry endpoint

The sweep deliberately never revives a terminal record — the provider refused that text, so re-running it costs a call and fails identically. Correct, right up until the premise changes (the text was edited, the provider changed its policy), and then there is no way back. `POST /v1/tenants/{tenant_id}/enrichments/retry` is the override: it clears the terminal markers so the next sweep picks those records up.

Under `/v1/tenants/`, and not by preference. The gateway publicly routes the `/v1/feedback-records` prefix and injects Hub credentials into it, so an endpoint there is internet-reachable and defended only by the ext_authz allowlist. A bulk operation that spends provider money on records already known to fail does not belong on the public side of that line — both tenant purges live under `/v1/tenants/` for the same reason.

**The cooldown (migration 024) is a security control, not a nicety.** Clearing is a request to spend a provider call on every permanently-failed record, so with no bound the endpoint is call → clear → sweep → all fail → call again, at one invocation per record per cycle from a single authenticated caller, and the Hub has no rate limiting anywhere else. One row per (tenant, enrichment) bounds it to *(terminal set / cooldown)*, and the window is stamped in the same statement as the delete so a caller cannot take a clear and dodge it.

Three behaviours worth reviewing specifically:

- a refused call **says so**, with the remaining wait. A caller who cannot tell "nothing to clear" from "rate limited" just calls again, which is what the cooldown exists to stop
- a **disabled** enrichment is refused *before* the cooldown is spent — clearing markers for a switched-off pipeline queues work the worker skips, so the records get re-marked and the caller has burned an hour for nothing
- every requested enrichment gets its **own** outcome, because a tenant can have one cleared, one cooling down and one switched off in the same call

Linear: https://linear.app/formbricks/issue/ENG-2376/auto-re-queue-failed-jobs

### Post-review hardening (95714dc / 2367927)

A full review + security pass over this diff fixed 10 findings before any human read it. The ones a reviewer should know changed the design:

- **The pending set now excludes records with an in-flight job of the enrichment's kind** (anti-join over the bounded in-flight set). This — not River uniqueness — is what prevents cross-lane double-enqueueing: the event path inserts with no unique options, so its jobs carry no key to collide with. It also stops an in-backoff head from starving the pending tail, and stops a target-language change having the tenant fan-out and the sweep pay for the same corpus twice.
- **The retry endpoint refuses (409) when `ENRICHMENT_RECONCILE_ENABLED=false`** — its 202 means "the next sweep picks these up", and with the kill switch off it would otherwise delete markers, burn the cooldown, and improve the failure counts while doing nothing.
- **The cooldown claim is atomic with the clear, under the tenant write lock** — decided inside the cooldown row's lock (8-way race test: exactly one wins), gated on the shared advisory lock like every other tenant-owned mutation (purge-race test binds it), with the marker-side tenant predicate for the index.
- **The sweep job carries `MaxAttempts: 1`** — a failed sweep dies and the next tick is the retry; with River's default 25, a failing sweep's backoff held the unique key and silently absorbed every subsequent tick.
- The reconcile lanes are named `*_reconcile` (not `*_backfill`) because `translations_backfill` was one transposition from the pre-existing `translation_backfills`; migration 025 adds partial pending indexes for sentiment/emotions (translation documented as deliberately unindexable); sweep config is one `EnrichmentSweepSpec`; `RetryAfterSeconds` is ceiled; the retry body uses the house `MaxBytesReader` idiom; the gate mapping is shared with the status service; the six env vars are documented in `.env.example`.

### Smoke-tested end to end (real stack, real browser)

Against a Hub database carrying **801 genuinely stranded records** (discarded jobs, zero runnable — the exact state ENG-2375's PR left behind as a fixture), with the PR-8845 banner watching from a real formbricks web app:

1. Reconciler enabled (15s interval): **one sweep enqueued 801, the provider was called exactly 801 times, 801 successes** — the anti-join provably prevented duplicates while the drain was in flight (concurrent sweeps enqueued 0).
2. `failed` went 400/401 → **0** for both enrichments and both directories, `unaccounted=0` throughout; terminal records correctly untouched. Steady-state sweeps: ~7–10ms.
3. **Retry endpoint live**: cleared 108+89 terminals (translation correctly `disabled`), immediate second call `cooling_down` with `retry_after_seconds: 3600`, the next sweep enqueued exactly the 197 cleared, and the tenant reached **1151/1151 done, zero failures of any kind**.
4. **The UI told the story**: "801 records pending" (spinner) → "Some feedback couldn't be enriched — 14 failed" → retry → "14 records pending" → banner gone, all caught up.
5. Bonus coverage: the run accidentally exercised **leader handover** — an older worker held River leadership so no sweeps fired (leader-only scheduling held across two competing clients); killing it transferred leadership and the periodic job started on the survivor.

### One definition of pending, everywhere (5c68634)

The reconciler skips terminally-failed records because a provider that refused a text structurally
will refuse it again. The settings-triggered per-tenant translation fan-out shared none of that
reasoning: it re-listed every untranslated record, terminal markers included. So the sweep would
carefully skip a record and the next target-language change would hand it another provider call —
1517 wasted jobs in a measured 3000-record run, and the terminal exclusion worth nothing in
practice. `translationBackfillSelectSQL` now carries the same exclusion the pending set does.

Non-terminal markers stay targets (those failed for reasons a retry can fix, which is the whole
distinction the taxonomy draws), and the join is per (record, enrichment), so a record the
sentiment provider refuses is still translatable. Both are pinned by test.

That same fan-out also enqueued onto the live `translations` queue, where a tenant's re-translated
history sits in front of records arriving right now — the starvation the reconcile lanes exist to
prevent. It moves to `translations_reconcile`, which is registered under the same condition as the
fan-out worker itself, so the lane exists whenever the fan-out can run, independent of the
reconciler's kill switch.

The three operator-run CLI backfill commands have the same two problems and are deliberately left
alone here — filed as ENG-2744, since retiring them may be the better answer than fixing them.

### Automated review round (6ebbb29)

Four fixes. The sweep's 5-minute timeout was decorative — `WorkerDefaults.Timeout` returns zero,
River reads that as "use `Config.JobTimeout`", `RIVER_JOB_TIMEOUT_SECONDS` defaults to 0, and River
reads *that* as its own one-minute default, so sweeps were cut at 60s and `MaxAttempts: 1` turned
each one into a discarded job and an error log per tick. A multi-enrichment retry that failed
part-way discarded the record of clears that had already burned their cooldowns. The endpoint
documented only 202 despite answering 400/401/409/413. And the pending-set test helper read a
capped cross-tenant page — adding a truncation guard turned it red immediately on a database that
had been used for other work, so it was passing on the accident of newest-first ordering.

**One finding is knowingly left open.** The in-flight anti-join is a read, not a lock, so an event
landing between the `SELECT` and the `InsertMany` still produces two jobs for one record. The
suggested fix — a shared cross-lane unique key — is the bug `enrichment_provider.go` documents at
length: `completed` sits unremovably in River's default unique-state set and swallows legitimate
re-enrichment on `A→cleared→A`. The alternative is a per-record lock at Work time, to buy back one
duplicate provider call on a record that happened to change during the milliseconds of one sweep —
and when the racing event *is* a content change, the supersession guards already drop the stale
result, so what survives is a wasted call rather than a wrong answer. The comment now states the
residual window instead of claiming the anti-join closes it.

### Metrics (e47a8d0)

The sweep runs unattended on the elected leader and the retry endpoint is called by another
service, so neither appears in request dashboards — until this commit the only evidence either had
run was a log line, and the only way to notice sweeps had stopped converging was to notice their
absence. The plan named these under PR5, which shipped with ENG-2375 before the reconciler existed,
so they fell between the two PRs.

```
hub_enrichment_reconcile_sweeps_total{outcome}
hub_enrichment_reconcile_duration_seconds{outcome}
hub_enrichment_reconcile_enqueued_total{enrichment}
hub_enrichment_retry_requests_total{enrichment,outcome}
```

Enqueued and duration together are the post-deploy question: is the backlog draining, and is a
sweep still finishing inside its own interval — a duration approaching
`ENRICHMENT_RECONCILE_INTERVAL_SECONDS` means ticks are about to overlap and be coalesced by the
job's uniqueness. The sweep records on **both** paths and before the error return, because a sweep
that failed part-way still enqueued what it enqueued; a test fails if that is narrowed back to the
success path. The retry counter uses the API's own outcome values rather than a generic
success/error, since `cleared` turning into `cooling_down` is the cost bound working — and because
`observability` cannot import `models`, a parity test binds the two enums (the normalizer rewrites
anything unrecognised to `error`, so a rename would otherwise silently mislabel that outcome
forever).

Also brings `EnrichmentReconcileConfig.Interval` and `.Depth` from 0% to full coverage. Both exist
only to survive misconfiguration, including the `TargetDepth` ceiling added during review.

### Second smoke run, on the final head (e47a8d0)

The first smoke run predated the lane change and the metrics, so both were re-verified live against
a real stack — Postgres, both binaries, a synthetic provider, and an OTLP sink capturing exported
telemetry.

**The lane move.** Setting `target_language` on a 60-record tenant fanned out **59 jobs onto
`translations_reconcile` and zero onto the live `translations` queue** — 59 not 60 because one
record carried a terminal marker and was correctly skipped. All 59 drained (58 completed, 1
cancelled), which is the part a unit test cannot show: the queue jobs are now inserted onto has
workers assigned, so they are not sitting there looking enqueued forever.

**The reconciler's own exclusion.** With 5 records matching a naive "not translated" count but 2 of
them terminal, the sweep enqueued exactly **3**. That is what makes convergence terminate.

**The terminal taxonomy, end to end.** The 1 cancelled job was a provider refusal: the worker
classified it terminal, cancelled rather than burning the retry budget, and wrote a durable marker
with `reason=refusal`. `POST /enrichments/retry` then cleared 2 markers (`202 cleared`), a second
call correctly refused (`cooling_down`, `retry_after_seconds: 3600`), and the next sweep re-picked
exactly those 2. Tenant finished at **60 eligible / 60 translated / 0 terminal** — also a live
demonstration that a terminal marker is not a life sentence. 64 jobs for 64 units of work; no
duplicate enqueueing.

**Metrics.** All four confirmed in exported OTLP payloads with correct labels —
`sweeps_total{outcome=success}`, `duration_seconds`, `enqueued_total{enrichment=translation}`, and
`retry_requests_total` carrying both `cleared` and `cooling_down`. The cardinality rule was checked
against the wire, not the source: **no `tenant_id` label key and no tenant id values appear
anywhere** in the exported telemetry.

## How should this be tested?

```bash
go test ./internal/... ./cmd/...
DATABASE_URL=... make tests
make lint && make migrate-validate
```

**The sweep's arithmetic is the part worth attacking**, since the whole rate-control story rests on it. `internal/service/enrichment_reconcile_test.go` pins: asks for exactly the remaining room (1000 target − 400 runnable = 600), skips a queue already at depth without even querying the pending set, never asks for a negative limit when a queue is past target, counts only what River actually inserted rather than duplicates it skipped, refuses to sweep with no inserter attached, and keeps sweeping the other enrichments when one query fails. Each was verified by breaking it and watching the suite go red — e.g. replacing the top-up with a fixed batch fails on "1000 target minus 400 already runnable", and dropping the explicit `ByState` fails the uniqueness assertion.

`tests/enrichment_reconcile_test.go` runs the pending set against real Postgres: a never-attempted record is pending, a retryable failure comes back, a `write_failed` comes back, a terminal one is excluded, and an enriched one is not pending. Verified by making the exclusion key on the reason instead of `terminal` — which strands every write failure — and by removing it entirely.

**The parity tests are the ones that caught real mistakes here.** Declaring a job kind without registering a worker, or declaring a backfill lane no worker is assigned to, both look exactly like enqueued work that never runs. Both now fail loudly.

**The retry endpoint** has five integration tests against real Postgres, each verified by injection: removing the cooldown check, spending the cooldown on a refused retry, and clearing non-terminal markers alongside terminal ones all turn the suite red. The cooldown test also proves the window *expires* rather than latching — a cooldown that never released would be a worse bug than none.

**By hand**, for Dhru's original repro: create a record with `SENTIMENT_PROVIDER`/`SENTIMENT_MODEL` unset, set them, and wait one interval. The record should enrich with no manual `backfill-classify` run. I have not yet automated this — see below.

### Required

- [x] Ran `make build`
- [x] If database schema changed: migration `024_enrichment_retry_cooldowns.sql`, goose annotations, `make migrate-validate` run

### Appreciated

- [x] If API changed: `openapi.yaml` updated for the retry endpoint; contract tests run
- [x] If API behavior changed: request/response shapes documented in the spec


internal/workers/enrichment_reconcile.go
```diff
@@ -0,0 +1,77 @@
...
+// Work runs the sweep.
+//
+// A failed sweep returns an error so River retries it, but the retry is not what makes the system
+// converge — the next scheduled tick is. That matters for how alarming a failure here should look:
+// one is noise, a run of them means coverage has stopped being guaranteed.
+func (w *EnrichmentReconcileWorker) Work(
+	ctx context.Context, _ *river.Job[service.EnrichmentReconcileArgs],
+) error {
+	ctx, cancel := context.WithTimeout(ctx, enrichmentReconcileTimeout)
+	defer cancel()
```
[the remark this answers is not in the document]
xernobyl: Good catch, and it was worse than "the constant is decorative" — thanks.

Fixed in 6ebbb29 by declaring `Timeout` on the worker. Confirming your chain: `WorkerDefaults.Timeout` returns zero → River falls back to `Config.JobTimeout` → `RIVER_JOB_TIMEOUT_SECONDS` is `env-default:"0"` and `cmd/worker/app.go` only sets `riverCfg.JobTimeout` when it is `> 0` → River reads the zero as `JobTimeoutDefault`, one minute. The `context.WithTimeout` in `Work` could not rescue it, a child context being unable to outlive its parent.

The consequence was the part I had missed: with `MaxAttempts: 1` the cancellation discarded the job and logged an error every tick, which is exactly the "a run of them means coverage has stopped being guaranteed" signal `Work`'s own doc comment describes — so the alarm would have been firing for a reason that was not the one it claims to indicate.

The peer `TenantTranslationBackfillWorker` already declared `Timeout`, so this was also an inconsistency between two workers doing the same kind of job. Added `TestEnrichmentReconcileWorkerTimeout`, which I verified fails when the method is deleted — the worker had no test file at all before, so nothing would have noticed.

internal/service/enrichment_retry_service.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	for _, enrichment := range requested {
+		result, retryErr := s.retryOne(ctx, normalizedTenantID, enrichment, settings)
+		if retryErr != nil {
+			return nil, retryErr
+		}
+
+		response.Results = append(response.Results, result)
+	}
```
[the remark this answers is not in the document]
xernobyl: Agreed, fixed in 6ebbb29 — though I took the logging half of your suggestion rather than returning the partial results.

The failure mode you describe is real and is worse than losing a response: `retryOne` deletes markers *and* stamps the cooldown, so a failure on the second enrichment leaves the first one cleared with its cooldown burned. The caller sees a 500, retries, gets `cooling_down`, and there is nothing anywhere connecting the two. That is the state the endpoint exists to make legible.

Where I diverged: returning `response.Results` alongside the error means a caller has to consume a partial success under a failure status, and the handler would have to decide which status code describes "one worked, one did not". A log line naming exactly what was already cleared answers the diagnostic question — "who caused this and what did it do" — without asking every client to model a half-success. If it turns out callers need to act on it programmatically rather than have a human read it, a per-enrichment `failed` outcome in the 202 body is the cleaner shape, and I would rather do that deliberately than fall into it via an error path.

openapi.yaml
```diff
@@ -1588,6 +1588,50 @@ paths:
...
+                    application/json:
+                        schema:
+                            $ref: '#/components/schemas/EnrichmentRetryInputBody'
+            responses:
+                '202':
+                    description: Accepted; see each enrichment's outcome.
+                    content:
+                        application/json:
+                            schema:
+                                $ref: '#/components/schemas/EnrichmentRetryOutputBody'
```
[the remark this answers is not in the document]
xernobyl: Fixed in 6ebbb29, and you found one instance of a broader omission — the endpoint documented only `202` while it can answer four other statuses, so I added all of them rather than just the `413`:

- `400` — unknown enrichment name, a body that is not a JSON object, or a missing tenant id
- `401` — missing or invalid API key
- `409` — the reconciler is disabled on this deployment, or a tenant data purge currently holds the tenant's write lock
- `413` — the `MaxBytesReader` cap you spotted
- plus the `default` error response the other endpoints in this spec carry

The `409` is the one I would least have wanted a generated client to miss, since it is not an error in the usual sense: it means the clear was deliberately refused and the caller should stop rather than retry.

Verified with `spectral lint` and the API contract tests are green.

tests/enrichment_retry_test.go
```diff
@@ -0,0 +1,377 @@
...
+	ids := map[uuid.UUID]bool{}
+
+	for _, target := range targets {
+		if target.TenantID == tenant {
+			ids[target.ID] = true
+		}
+	}
+
+	return ids
+}
```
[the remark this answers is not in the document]
xernobyl: You were right, and not hypothetically — thanks for pushing on this one.

I initially added a truncation guard to *prove* the concern was theoretical, expecting it to stay green. It went red on the first run: the database I was working against already held ≥500 pending records, so the page genuinely was truncated and the test had been passing only on the accident of the query's newest-first ordering putting the seeded records at the top. On a fresh CI database it passes for a reason that is a property of the suite, not of the code.

Fixed in 6ebbb29. I kept the real `ListPendingEnrichment` rather than substituting a tenant-scoped test query, because what these tests assert is that *the sweep* sees or stops seeing a record — a parallel query written for the test could agree with the assertion while disagreeing with the sweep, which is the failure mode this whole PR exists to remove. Instead `pendingLimit` is now set far above anything the suite creates, and the guard stays: if it is ever exceeded the failure says "this page is truncated, raise the limit" rather than blaming a record that was simply pushed off the end.

Deliberately not adding a tenant parameter to `ListPendingEnrichment` — it is cross-tenant by design, and widening a production signature to suit a test would be the wrong trade.

internal/repository/enrichment_reconcile_repository.go
```diff
@@ -0,0 +1,228 @@
...
+		-- build over that small set.
+		SELECT (j.args->>'feedback_record_id') AS record_id
+		FROM river_job j
+		WHERE j.kind = '` + jobKind + `' AND j.state IN ` + pendingInFlightStates + `
+	) inflight ON inflight.record_id = fr.id::text
+	WHERE ` + enrichmentEligibleText + `
+		AND ` + gate + `
+		AND ` + notDone + `
+		AND f.feedback_record_id IS NULL
+		AND inflight.record_id IS NULL
```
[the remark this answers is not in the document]
xernobyl: The diagnosis is correct and I have left the behaviour as-is deliberately. 6ebbb29 rewrites the comment here to say so, because it previously claimed the anti-join "actually keeps the sweep from double-enqueueing", which oversells a read as a lock.

You are right that `ListPendingEnrichment` and `InsertMany` are not atomic, and that an event landing between them yields two jobs for one record.

On the suggested fix — a shared cross-lane uniqueness key is the one option that is not available. `enrichment_provider.go` documents at length why the event path inserts with *no* `UniqueOpts`: River's default unique state set includes `completed`, which cannot be removed, so hash-dedupe silently swallowed legitimate re-enrichment whenever content transitioned away and back inside the dedupe window (`A → cleared → A` left the enrichment NULL; `A → "" → B → ""` kept B's result on an empty record). Giving these two lanes a shared key reintroduces exactly that bug, and it is a correctness regression rather than a cost one.

That leaves a per-record lock at Work time. What it buys back is one duplicate provider call on a record that changed during the milliseconds between one sweep's SELECT and its InsertMany. And when the racing event *is* a content change — the common case, since `FieldsChangedFrom` means every event is a real transition — the supersession guards (`ErrClassificationSuperseded` / `ErrTranslationSuperseded`) already drop the stale result, so what survives is a wasted call, not a wrong answer. The genuinely wasteful case is narrower still: a record created inside that window, where both jobs see identical content.

That is a real cost and I am not dismissing it, but a distributed per-record lock on every enrichment write path is a large amount of machinery relative to it, and it would sit on the hot event path to fix a problem the reconciler introduces. The level-triggered design accepts a narrow duplicate window by construction — same property a Kubernetes controller has — and the anti-join reduces it from "every sweep re-enqueues everything in flight" to "a millisecond-wide race". Happy to be overruled by a human reviewer if they weigh it differently; it would be its own ticket rather than a change to this one.

internal/repository/enrichment_reconcile_repository.go
```diff
@@ -0,0 +1,239 @@
...
+//
+// "Runnable" is the set that has not finished and is not being retried into the future:
+// available, pending, running and scheduled. Retryable is deliberately EXCLUDED — a job waiting
+// out its backoff will run again on its own, so counting it would have the reconciler treat a
+// queue as full while nothing is draining, and a stuck backoff would stall the sweep indefinitely.
+const countRunnableByQueueSQL = `
+	SELECT queue, count(*)
+	FROM river_job
+	WHERE queue = ANY($1)
+	  AND state IN ('available', 'pending', 'running', 'scheduled')
```
Dhruwang: **`retryable` is excluded here but *included* in the pending set's in-flight states, and the combination is not bounded by `TargetDepth`.**

`countRunnableByQueueSQL` counts `available, pending, running, scheduled`; `pendingInFlightStates` (line 73) counts `available, pending, retryable, running, scheduled`. Each choice is well argued on its own, but together they mean retryable work is invisible to the depth control *and* its records are invisible to the pending query — so a tick sees depth ~0 and enqueues a full `TargetDepth` of **different** records on top of the backoff pile.

Concretely, with `SENTIMENT_MAX_ATTEMPTS=8` (allowed; env-default is 3), `INTERVAL_SECONDS=300`, `TARGET_DEPTH=1000`, a large backlog, and the provider returning 5xx: River's default policy backs off `attempt^4` seconds, so a job sits `retryable` for 1+16+81+256+625+1296+2401 ≈ 78 min. That is ~15 ticks inside one retry window ⇒ ~15,000 rows on `sentiments_reconcile` and ~120,000 provider calls, against the documented bound of 1,000.

Harmless at the default `MAX_ATTEMPTS=3` (retry window ~17s, well under the interval), so this is a config-reachable hole rather than a today-bug. Worth either clamping `room` against a count that includes `retryable`, or documenting that the `TargetDepth` bound holds only while the retry window is shorter than the interval.

internal/service/enrichment_retry_service.go
```diff
@@ -0,0 +1,283 @@
...
+//
+// Every enrichment gets an outcome, including the ones that were refused. A caller that cannot
+// tell "there was nothing to clear" from "you are being rate limited" will simply call again,
+// which is the behaviour the cooldown exists to stop.
+func (s *EnrichmentRetryService) Retry(
+	ctx context.Context, tenantID string, enrichments []string,
+) (*models.EnrichmentRetryResponse, error) {
+	// Refused BEFORE validation spends anything: with the reconciler off there is no sweep to pick
+	// the cleared records up, so a 202 here would be a lie the caller pays a cooldown for.
+	if !s.reconcileEnabled {
```
Dhruwang, line 137: **This guard reads hub-api's copy of `ENRICHMENT_RECONCILE_ENABLED`, but the sweep it describes runs in hub-worker.**

`ReconcileEnabled: cfg.EnrichmentReconcile.Enabled` is resolved in `cmd/api/app.go`, and nothing correlates the two processes' values. `charts/hub/templates/worker-deployment.yaml` reads `worker.extraEnv` / `worker.extraEnvFrom`; `deployment.yaml` does not — so the natural way to switch off a worker-only feature sets it on the worker alone.

In that state hub-api still has the `env-default:"true"`, so `POST /v1/tenants/{id}/enrichments/retry` returns `202 cleared`: the terminal markers are deleted, the (tenant, enrichment) cooldown is burned, the records drop out of the status endpoint's `failed_terminal` count, and no sweep will ever pick them up — the exact "coverage would silently look BETTER while nothing happened" outcome the comment above says this prevents. The reverse (set on hub-api only) hard-409s every retry while the worker is sweeping normally.

Nothing here is wrong for a single-config deployment; the gap is that the invariant "both processes agree on this flag" is load-bearing and neither enforced nor detectable. A line in the env-var docs saying it must be set identically on both would close it cheaply.

internal/service/enrichment_reconcile.go
```diff
@@ -0,0 +1,258 @@
...
+
+			continue
+		}
+
+		queues = append(queues, queue)
+	}
+
+	depths, err := s.repo.CountRunnableByQueue(ctx, queues)
+	if err != nil {
+		return result, fmt.Errorf("read backfill queue depths: %w", err)
```
Dhruwang: **The `ErrUnknownEnrichment` entries collected just above are dropped by this early return.**

The loop at lines 126-139 appends to `errs` specifically so an enrichment that is enabled but missing from `ReconcileQueueFor` fails loudly instead of silently vanishing from the sweep (see the comment at 129-133). This `return` discards `errs` entirely.

So: add a fourth enrichment to the spec list but not to `ReconcileQueueFor`, and on any tick where the depth query also fails (connection reset, statement timeout on `river_job`) the worker logs only `read backfill queue depths: ...`. The operator reads a transient DB error, retries, and the permanent misconfiguration — the one that leaves an enrichment never swept — is never surfaced.

`errors.Join(append(errs, fmt.Errorf("read backfill queue depths: %w", err))...)` keeps both.

tests/enrichment_reconcile_test.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+
+	assert.False(t, ids2[inflight.ID], "a record with an in-flight job must not be re-enqueued")
+	assert.True(t, ids2[finished.ID], "a finished job is not in flight; the record is pending again")
+	assert.True(t, ids[transient.ID], "a retryable provider failure must come back as pending")
+	assert.True(t, ids[writeFailed.ID], "a failed write is retryable too, and must come back as pending")
+	assert.False(t, ids[terminal.ID], "a terminal failure must be excluded or the sweep never ends")
+	assert.False(t, ids[doneRec.ID], "an enriched record must not be pending")
+	assert.Len(t, ids, 3, "exactly the three pending records, no more")
```
Dhruwang, line 118: **This asserts on a capped cross-tenant page with no truncation guard.**

`ListPendingEnrichment` is global with `LIMIT 500` and `ORDER BY fr.collected_at DESC, fr.id DESC`, and the per-tenant filter is applied in Go *after* the page comes back. So this length assertion depends on the seeded records landing inside the newest 500 rows of a shared database, not on the query under test.

Anything that puts ≥500 un-enriched text records at or after these seeds' `collected_at` — the 3000-record fixture the PR description mentions, or a second `tests` binary against the same DB — pushes the five seeds off the page, `ids` comes back empty, and line 112 fails with nothing indicating the query was fine and only the page was full. It fails in the other direction too: a change that broke the terminal exclusion could still pass here if `terminal` happened to be off the page.

The PR description says a truncation guard was added for exactly this ("adding a truncation guard turned it red immediately"), but I don't see one on this head — `require.Less(t, len(got), 500)` before the assertions would restore it.

Dhruwang (commented):
Correctness-focused pass over the diff. Build, `go vet` and `go test ./internal/... ./cmd/...` are green locally (integration tests need a DB, not run).

I checked and could not fault: the retry SQL's atomic claim (the `gate` CTE holds a volatile function so Postgres materializes it once; the data-modifying `claim` CTE is always materialized so the `EXISTS` ordering is deterministic; `fr.tenant_id` stays the boundary and the PK makes the unqualified `DELETE` safe), the in-flight anti-join (no row fan-out survives `IS NULL`, and `fr.id::text` matches Go's UUID marshalling), River 0.39 genuinely honouring `UniqueOpts` on `InsertMany` (the unique key is computed in `insertManyShared`, and `InFlightUniqueStates()` satisfies `requiredV3states`), migration 025's partial predicates being provably implied by the queries (`negate_clause` De-Morgans the emotions `NOT (… OR …)` into the index's `IS NULL AND IS NULL`), pgx v5's `interval` ↔ `time.Duration` mapping, and the reconcile-lane queue registration conditions matching the sweep's spec conditions in all three cases.

Four comments below — one config-reachable bound, one cross-process config coupling, one dropped error, one order-dependent test. None block on their own.

Unrelated nit: the description's config section still says `{TRANSLATION,SENTIMENT,EMOTIONS}_BACKFILL_MAX_CONCURRENT`; the code and `.env.example` both use `*_RECONCILE_MAX_CONCURRENT`.

internal/repository/enrichment_reconcile_repository.go
xernobyl: Confirmed and fixed in 2608e44. This was the best finding of the round — the interaction is real and I had reasoned about each half in isolation without ever putting them together.

The two sets now share one definition (`pendingInFlightStates`), so the depth control counts exactly what the pending set treats as already handled. They have to agree or the bound means nothing.

On the comment that defended the exclusion — it argued that counting `retryable` would let a stuck backoff stall the sweep. I no longer think that holds. `retryable` is a bounded state: a job either succeeds or exhausts `MaxAttempts` and leaves for discarded/cancelled, writing its failure marker. And pausing is the *correct* behaviour while a lane is full of backing-off work — the provider is failing, and topping up another `TargetDepth` of records against a failing provider spends money to make the backlog worse. The event path is unaffected either way, so live enrichment still flows.

Your arithmetic is the part that made this concrete: ~15 ticks inside one retry window at `MAX_ATTEMPTS=8`, ~15,000 rows against a documented bound of 1,000. Agreed it is config-reachable rather than a default-config bug, but "the bound holds only if you don't raise a documented env var" is not a bound.

Added `TestQueueDepthCountsRetryableJobs`, which asserts all five in-flight states count and that finished work does not (counting `completed` would wedge the lane at its target forever). Verified by injection: restoring the four-state list turns it red.

internal/service/enrichment_retry_service.go
xernobyl: Correct, and fixed as documentation in 2608e44 rather than as a mechanism.

You traced it exactly: the flag gates a refusal in hub-api, the sweep it refuses on behalf of runs in hub-worker, and nothing correlates the two. The asymmetric case you name is the dangerous one — `worker.extraEnv` is the natural place to switch off a worker-only feature, and doing only that leaves the API returning `202 cleared`, deleting markers, burning the tenant's cooldown, and dropping those records out of `failed_terminal` for a sweep that will never come. That is precisely the outcome the guard was written to prevent, arrived at from the other direction.

I went with your suggested fix rather than a mechanism. The alternatives — having the worker publish its state to a table the API reads, or moving the decision into the sweep — buy correctness for a misconfiguration at the cost of a new cross-process dependency on the retry path, and the Hub has no such coupling anywhere today. A stated invariant is the proportionate answer.

It is now on the env-vars reference (both in the table and in prose explaining what each half-configured state does) and in `.env.example`. Notably `.env.example` already carried a "configure identically in the API and worker processes" note for the translation vars, so this now matches an established pattern rather than inventing one.

internal/service/enrichment_reconcile.go
xernobyl: Confirmed and fixed in 2608e44 — a straightforward bug, and your suggested resolution is what I used.

The `errs` slice exists precisely so an enrichment enabled but absent from `ReconcileQueueFor` fails loudly instead of vanishing from the sweep, and then the depth-query early return threw it away. The asymmetry is what makes it bite: the wiring mistake is permanent and the database error is transient, so the operator sees the one they can retry and never learns about the one that is quietly starving an enrichment.

Now `errors.Join`s both, with a comment saying why the permanent error must not be swallowed by the transient one.

Added `TestSweepKeepsWiringErrorsWhenTheDepthQueryFails`, which asserts the returned error matches both `ErrUnknownEnrichment` and the depth failure. Verified by injection: restoring the plain early return turns it red on the wiring assertion.

tests/enrichment_reconcile_test.go
xernobyl: You are right, and the detail that stings is that the PR description was accurate about the wrong file.

The guard went onto `pendingIDs` in `tests/enrichment_retry_test.go` — where it immediately turned red and proved the concern was not theoretical — and I never applied it to `tests/enrichment_reconcile_test.go`, which reads the same capped cross-tenant query with the same in-Go tenant filter. Grepping for the helper I had guarded would only ever have found the site that was already safe; the shape is what needed sweeping for, and I did not.

Fixed in 2608e44: both calls here now use a shared `reconcilePageLimit`, set far above what the suite creates, with `require.Less` after each fetch so a full page fails as "this page is truncated, raise the limit" rather than as an assertion about a record that was simply pushed off the end. Same limit and same guard as the retry tests, with a comment on each pointing at the other.

Your second direction is the sharper one and worth recording: it fails silently in the *other* direction too — a regression that broke the terminal exclusion could still pass here if `terminal` happened to fall off the page. The guard closes both.

Verified by injection: dropping the limit below the seeded set turns it red with the truncation message.

xernobyl:
Thanks Dhru — all four are fixed in 2608e44, and the description nit is corrected (it now reads `*_RECONCILE_MAX_CONCURRENT`, matching the code and `.env.example`).

Two notes on the replies in the threads.

The depth-count one is the finding of the round. I had reasoned about each state set on its own and never put them side by side, so the interaction was invisible to me. Worth flagging that fixing it meant reversing a decision the code explicitly defended: the old comment claimed counting `retryable` would let a stuck backoff stall the sweep. I no longer think that holds — `retryable` is bounded, since a job either succeeds or exhausts `MaxAttempts` and leaves for discarded/cancelled — and pausing while a lane is full of backing-off work is the behaviour we want, not a stall. If you read that differently, it is the one change here I would most want a second opinion on.

I also resolved the four threads after replying. If you would rather check the fixes yourself first, say so and I will reopen them — I should probably have left that to you.

Thanks too for the "could not fault" list. Several of those (the `gate` CTE materialization, `requiredV3states`, the De-Morgan on the emotions partial index) are exactly the things I would not have been able to prove on my own, and knowing they were checked is worth as much as the findings.

Dhruwang (approved):

PR #132, opened 2026-08-26
fix(ci): sign release image digest once

BhagyaAmarasinghe:
## Summary

- sign the canonical versioned GHCR image reference once per release
- retry Cosign up to three times when transient Rekor failures abort a signing invocation
- allow the dependent ECR and release-verification jobs to run

## Context

The 0.8.5 release build pushed one multi-arch digest under four tags. Two signing invocations failed when Rekor returned 409 equivalent-entry conflicts after a long request, while fresh invocations for the same digest succeeded. The workflow now signs the digest once and retries with a fresh keyless invocation when signing fails.

Failed run: https://github.com/formbricks/hub/actions/runs/32244009436
Upstream Cosign issue: https://github.com/sigstore/cosign/issues/4711

## Test evidence

- Ruby YAML syntax parse and signing-step lookup
- Bash syntax check of the extracted signing step
- transient-failure dry run: first attempt failed, second succeeded
- permanent-failure dry run: exactly three attempts, exit status 1
- git diff --check

### DIFF
--- a/.github/workflows/release-docker-ghcr.yml
+++ b/.github/workflows/release-docker-ghcr.yml
@@ -135,11 +135,11 @@ jobs:
 
       - name: Sign image with cosign
         env:
-          TAGS: ${{ steps.tags.outputs.tags }}
+          IMAGE_REF: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ steps.extract_version.outputs.VERSION }}
           DIGEST: ${{ steps.build.outputs.digest }}
         run: |
           set -euo pipefail
-          echo "${TAGS}" | xargs -I {} cosign sign --yes "{}@${DIGEST}"
+          cosign sign --yes "${IMAGE_REF}@${DIGEST}"
 
       - name: Output build summary
         env:
--- a/.github/workflows/release-docker-ghcr.yml
+++ b/.github/workflows/release-docker-ghcr.yml
@@ -139,7 +139,20 @@ jobs:
           DIGEST: ${{ steps.build.outputs.digest }}
         run: |
           set -euo pipefail
-          cosign sign --yes "${IMAGE_REF}@${DIGEST}"
+
+          for attempt in 1 2 3; do
+            if cosign sign --yes "${IMAGE_REF}@${DIGEST}"; then
+              exit 0
+            fi
+
+            if [[ "${attempt}" -lt 3 ]]; then
+              echo "cosign sign failed (attempt ${attempt}/3); retrying in 10 seconds..."
+              sleep 10
+            fi
+          done
+
+          echo "cosign sign failed after 3 attempts"
+          exit 1
 
       - name: Output build summary
         env:


xernobyl (commented):
Signing once per digest is the right call 👍 — signatures are digest-scoped, so the one signature covers every tag, and `cosign verify` on `:latest` / `:0.8` etc. still works. Change looks correct (ran actionlint + dry-ran the step logic for stable/prerelease versions; `IMAGE_REF` always matches the first pushed tag).

One thing I'd flag though: I dug through the [failed run](https://github.com/formbricks/hub/actions/runs/32244009436) and I don't think the per-tag loop was the actual root cause, so this likely reduces the odds a lot but may not fully fix it:

- invocation 1 (`:0.8.5` — the exact ref we now sign) failed with `createLogEntryConflict`
- invocations 2 and 4 (`:0.8`, `:latest`) signed the *same digest afterwards* and succeeded

So Rekor doesn't dedupe by digest across invocations — each keyless sign uses a fresh ephemeral key, so every entry is unique, and an "equivalent entry" can only come from the same invocation. The failing invocations also stalled ~30–40s before erroring vs ~4s for the successful ones, which looks like the Rekor client retrying a POST that had already committed server-side: timeout → retry → 409 → fatal (cosign tlog upload is known to be non-idempotent: sigstore/cosign#3356). Under the same Rekor flakiness, the single invocation can still 409 and skip ECR + verify again.

Suggestion: wrap it in a plain retry. A fresh invocation mints a fresh key, so a previous 409 can't block it — the failed run itself demonstrates this (invocation 2 succeeded seconds after invocation 1's 409):

```bash
for attempt in 1 2 3; do
  cosign sign --yes "${IMAGE_REF}@${DIGEST}" && exit 0
  echo "cosign sign failed (attempt ${attempt}), retrying..."
  sleep 10
done
exit 1
```

(I'd avoid just swallowing the 409 — in cosign v3 the tlog proof is embedded in the bundle before the registry attach, so aborting on the 409 means no signature gets attached to the image at all. Bumping cosign might help too — the installer pin gives 3.0.6, 3.1.1 is current — but I couldn't find a confirmed upstream fix for the 409 handling, so the retry feels like the reliable part.)

nit: the tags step hardcodes `ghcr.io/${IMAGE_NAME}` while the sign step now uses `${{ env.REGISTRY }}` — same value today, might be worth unifying at some point.

fwiw I double-checked the registry: 0.8.5's digest (`sha256:9f4c10…`) did get its bundle from the invocations that succeeded, so the shipped image is signed — the damage was just the duplicate tlog submissions + the skipped ECR/verify jobs.

xernobyl (approved):

BhagyaAmarasinghe:
@xernobyl Addressed the reliability concern in 2eadc06: the canonical digest signing now retries up to three fresh Cosign invocations, waiting 10 seconds between failures, and still fails closed after the third attempt. I dry-ran both paths: transient failure recovered on attempt 2; permanent failure stopped after exactly 3 attempts with exit status 1. I left the registry-variable consistency nit unchanged as non-blocking.

xernobyl:
Post-merge follow-up — the retry held up on the first real release 👍

0.8.6 went out ~4 min after this merged, so it's already production-proven: signing succeeded on attempt 1 in ~6s, against the 30–40s stalls that preceded the 409s in the 0.8.5 run.

I also went and checked the digest-scoped claim against the live registry rather than assuming it. One signature now covers every tag:

| release | signing code | signatures on the digest |
|---|---|---|
| 0.8.4 | old per-tag loop, all 4 succeeded | 4 |
| 0.8.5 | old per-tag loop, 2 of 4 409'd | 2 |
| 0.8.6 | this PR | 1 |

And all four 0.8.6 tags verify against that single signature — `cosign verify` exits 0 on `:0.8.6`, `:0.8`, `:0` and `:latest`, all resolving to `ghcr.io/formbricks/hub:sha256-5f9afc8f….sig`. So it really is keyed by digest and tag-independent, and we've stopped submitting 3 redundant tlog entries per release as a bonus.

Then dry-ran the merged step with a stubbed cosign to confirm the loop itself behaves:

| cosign fails first N | exit | invocations | backoffs |
|---|---|---|---|
| 0 | 0 | 1 | 0 |
| 1 | 0 | 2 | 1 |
| 2 | 0 | 3 | 2 |
| 3+ | 1 | 3 | 2 |

No wasted attempt after a success, caps at exactly 3, no trailing sleep before the failure exit, and the argv stays digest-pinned (`sign --yes <ref>@<digest>`) on every attempt. One extra case I poked at: if `steps.build.outputs.digest` ever comes back empty, cosign fails on ref parsing and exits 1 — so it fails closed rather than quietly skipping signing, it just burns 3 attempts and 20s getting there.

Nothing to change, just closing the loop since my earlier comment wasn't confident the retry would be enough — turns out it was. 🙏

fwiw one tiny thing in the same spirit as the nit you left: the `:${VERSION}` in `IMAGE_REF` is inert, since cosign resolves `repo:tag@digest` purely by digest. Harmless, and it does keep the log line readable, so no reason to touch it.

PR #134, opened 2026-09-01
fix: reject unstorable feedback-record metadata as 400 instead of 500

xernobyl:
## What does this PR do?

Fixes ENG-2745 (https://linear.app/formbricks/issue/ENG-2745/hub-returns-500-instead-of-400-for-null-bytes-in-feedback-record)

`metadata` was the only field on the feedback-record create/update requests with no content validation — and it couldn't have had it: `no_null_bytes` skips any non-string kind, and `json.RawMessage` is a `[]byte`. So input the OpenAPI spec already declares invalid reached Postgres verbatim and failed the jsonb write as an unmapped 500.

Postgres refuses three things in jsonb that pass every upstream check, and this PR maps all of them to a 400 naming `metadata`:

- **NUL bytes** and **unpaired UTF-16 surrogates** (in keys or values) — a new `storable_json` validator tag scans the raw bytes on both request structs. Raw bytes on purpose: `encoding/json` rewrites both cases on decode (NUL escape → real NUL, lone surrogate → U+FFFD), so an unmarshalled value looks storable while the wire bytes are not.
- **numbers outside numeric range** (`{"n":1e1000000}` — 15 bytes of valid JSON) — invisible to any byte scan, so SQLSTATE 22003 is mapped at the repository on both write paths. Attribution to `metadata` is sound: it's the only place a numeric can originate on these queries (`value_number` is float8, and every float64 fits float8).

Also rewrites the three `metadata` descriptions in `openapi.yaml` (what belongs in it, snake_case keys, PATCH replaces wholesale, jsonb normalizes key order/number formatting — the old text said little beyond "additional context").

Before / after for the same request:

```json
// before                                           // after
{                                                   {
  "status": 500,                                      "status": 400,
  "code": "internal_server_error",                    "code": "validation",
  "detail": "An unexpected error occurred"            "invalid_params": [{
}                                                       "name": "metadata",
                                                        "reason": "must not contain NULL bytes or unpaired UTF-16 surrogates"
                                                      }]
                                                    }
```

## How should this be tested?

Against a local stack (`docker run -d -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=test_db -p 5432:5432 pgvector/pgvector:pg18`, `.env` with `DATABASE_URL=postgres://postgres:postgres@localhost:5432/test_db?sslmode=disable` and `API_KEY`, then `make init-db && make run-api`):

- `POST /v1/feedback-records` with `"metadata":{"note":"badvalue"}` → 400, `invalid_params` names `metadata` (was 500, `SQLSTATE 22P05` in the log)
- same with `"bad\ud83dvalue"`, a NUL in a **key**, and `{"n":1e1000000}` → 400 each; `PATCH` behaves identically
- a control: the same NUL in `value_text` still 400s as before, and ordinary metadata (nested objects, emoji, valid surrogate pairs, 200 keys, a 20k value) still stores and round-trips
- `tests/feedback_record_metadata_validation_test.go` drives all three rejection classes through the full HTTP stack against the real database
- differential check: 8,000 generated escape-soup JSON documents — the scanner's verdict matched Postgres's own `::jsonb` verdict on every one, in both directions

### Required

- [x] Ran `make build`
- [ ] If database schema changed: added migration in `migrations/` with goose annotations and ran `make migrate-validate` (no schema change)

### Appreciated

- [ ] Updated docs in `docs/` if changes were necessary (openapi.yaml carries the doc change)

---

> [!NOTE]
> **AI model used** — `claude-fable-5`, reasoning effort `unknown`.

### DIFF
--- a/internal/api/validation/validation.go
+++ b/internal/api/validation/validation.go
@@ -104,6 +104,10 @@ func init() {
 		slog.Error("Failed to register no_null_bytes validator", "error", err)
 	}
 
+	if err := validate.RegisterValidation(storableJSONTag, validateStorableJSON); err != nil {
+		slog.Error("Failed to register storable_json validator", "error", err)
+	}
+
 	// Element validators for the repeatable enum filters, applied via `dive`. They are a second
 	// gate behind registerEnumSliceTypes, not a replacement: see the comment there for why the
 	// decoder alone cannot be trusted to have run.
@@ -390,6 +394,8 @@ func FormatFieldError(fieldErr validator.FieldError) string {
 		return "must be in RFC3339 (ISO 8601) format"
 	case "no_null_bytes":
 		return "must not contain NULL bytes"
+	case storableJSONTag:
+		return "must not contain NULL bytes or unpaired UTF-16 surrogates"
 	case "http_url":
 		return "must be a valid HTTP or HTTPS URL"
 	case "url":
--- a/internal/api/validation/validation.go
+++ b/internal/api/validation/validation.go
@@ -395,7 +395,7 @@ func FormatFieldError(fieldErr validator.FieldError) string {
 	case "no_null_bytes":
 		return "must not contain NULL bytes"
 	case storableJSONTag:
-		return "must not contain NULL bytes or unpaired UTF-16 surrogates"
+		return "must contain valid UTF-8 and no NULL bytes or unpaired UTF-16 surrogates"
 	case "http_url":
 		return "must be a valid HTTP or HTTPS URL"
 	case "url":
--- a/internal/repository/feedback_records_repository.go
+++ b/internal/repository/feedback_records_repository.go
@@ -21,6 +21,17 @@ import (
 // constraint violations (23505 unique_violation).
 const uniqueViolationSQLState = "23505"
 
+// numericOverflowSQLState (22003 numeric_value_out_of_range) fires when jsonb metadata carries a
+// number that does not fit Postgres numeric — e.g. `{"n":1e1000000}`, fifteen bytes of perfectly
+// valid JSON. The storable_json request validator cannot see this class: it is a range property of
+// the parsed number, not of the bytes. Metadata is the only place it can originate on these
+// queries — value_number is float8, and every Go float64 fits float8 — so the error is mapped to a
+// metadata validation failure rather than surfacing as an unmapped 500 (ENG-2745).
+const numericOverflowSQLState = "22003"
+
+// numericOverflowMessage is the invalid_params reason for the mapping above.
+const numericOverflowMessage = "contains a number outside the storable range"
+
 // FeedbackRecordsRepository handles data access for feedback records.
 type FeedbackRecordsRepository struct {
 	db *pgxpool.Pool
@@ -137,6 +148,10 @@ func (r *FeedbackRecordsRepository) Create(ctx context.Context, req *models.Crea
 		var pgErr *pgconn.PgError
 		if errors.As(err, &pgErr) && pgErr.Code == uniqueViolationSQLState {
 			return nil, huberrors.NewConflictError("a feedback record with this tenant_id, submission_id, and field_id already exists")
+		}
+
+		if errors.As(err, &pgErr) && pgErr.Code == numericOverflowSQLState {
+			return nil, huberrors.NewValidationError("metadata", numericOverflowMessage)
 		}
 
 		if errors.Is(err, pgx.ErrNoRows) {
@@ -945,6 +960,11 @@ func (r *FeedbackRecordsRepository) Update(
 		if scanErr != nil {
 			if errors.Is(scanErr, pgx.ErrNoRows) {
 				return huberrors.NewNotFoundError("feedback record", "feedback record not found")
+			}
+
+			var pgErr *pgconn.PgError
+			if errors.As(scanErr, &pgErr) && pgErr.Code == numericOverflowSQLState {
+				return huberrors.NewValidationError("metadata", numericOverflowMessage)
 			}
 
 			return fmt.Errorf("failed to update feedback record: %w", scanErr)
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3061,9 +3061,11 @@ components:
                         a JavaScript client will still lose precision above 2^53 when it parses the
                         response.
 
-                        Nothing here is redacted or sent anywhere else — unlike `value_text`, which
-                        the enrichment pipelines also ship to LLM and embedding providers. Erasure
-                        covers metadata, with one precondition worth knowing:
+                        Nothing here is redacted. Metadata is **not** sent to the LLM or embedding
+                        enrichment providers, unlike `value_text` — but it **is** included in the
+                        `feedback_record.created` and `feedback_record.updated` webhook payloads, so
+                        it reaches whatever URL a tenant has configured. Erasure covers metadata,
+                        with one precondition worth knowing:
                         `DELETE /v1/feedback-records` matches on `user_id` and hard-deletes the whole
                         row, metadata included — so a record written **without** a `user_id` cannot
                         be reached that way. Set `user_id` on anything you may later have to erase.
@@ -3286,17 +3288,20 @@ components:
                     description: ISO language code. NULL bytes not allowed.
                     pattern: '^[^\x00]*$'
                 metadata:
+                    type: [object, array, string, number, boolean, "null"]
+                    # Required alongside `array` in the union; any element is valid, since the union
+                    # exists to describe legacy rows rather than to constrain new writes.
+                    items: {}
                     description: >-
                         Arbitrary context stored with this record, returned with equivalent values
                         (key order is normalized, and so are number exponents — `1e2` reads back as
                         `100`). Omitted when the record has no metadata; a record explicitly stored
                         with a JSON `null` returns `null`.
 
-                        Deliberately untyped: an object is the supported shape and what the request
-                        schemas accept, but the column stores whatever JSON was written, so a record
-                        created before that convention can return any JSON value. Constraining this
-                        to an object would make a generated client reject such a record.
-                    additionalProperties: {}
+                        An object is the supported shape and the only one the request schemas accept.
+                        The wider union is for reading: the column stores whatever JSON was written,
+                        so a record created before that convention can return an array or a scalar,
+                        and an object-only response schema would make a generated client reject it.
                 sentiment:
                     $ref: '#/components/schemas/SentimentValue'
                     description: Sentiment polarity inferred from value_text (sentiment enrichment). Read-only; absent until the record is enriched.
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3047,7 +3047,21 @@ components:
                     pattern: '^[^\x00]*$'
                 metadata:
                     type: object
-                    description: User agent, device, location, referrer, tags, etc. NULL bytes (\x00 or \u0000) are not allowed in JSON keys or values.
+                    description: >-
+                        Arbitrary context for this record, stored as JSON and returned verbatim: the
+                        dimensions you want to slice a dashboard by, such as channel, device, browser,
+                        OS, country, referrer, campaign, plan or tags. Values may be strings, numbers,
+                        booleans, null, or nested objects and arrays.
+
+                        Use snake_case keys and keep them stable across records, since a key is what a
+                        dashboard groups by: `customer_tier`, not a mix of `customerTier` and `tier`.
+                        Numbers keep full precision here (unlike `value_number`, which is a float64),
+                        so a large integer id round-trips exactly.
+
+                        Do not put personal data in metadata unless you mean to retain it — it is not
+                        redacted, and unlike an answer it is duplicated onto every record of a
+                        submission. Rejected with 400: NULL bytes and unpaired UTF-16 surrogates, in
+                        keys or values, because neither can be stored as JSON.
                     additionalProperties: {}
                 source_id:
                     type: [string, "null"]
@@ -3264,7 +3278,9 @@ components:
                     pattern: '^[^\x00]*$'
                 metadata:
                     type: object
-                    description: Additional context
+                    description: >-
+                        Arbitrary context stored with this record, returned exactly as it was sent.
+                        Absent when the record carries none.
                     additionalProperties: {}
                 sentiment:
                     $ref: '#/components/schemas/SentimentValue'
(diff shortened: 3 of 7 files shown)


internal/api/validation/validation.go
```diff
@@ -390,6 +394,8 @@ func FormatFieldError(fieldErr validator.FieldError) string {
 		return "must be in RFC3339 (ISO 8601) format"
 	case "no_null_bytes":
 		return "must not contain NULL bytes"
+	case storableJSONTag:
+		return "must not contain NULL bytes or unpaired UTF-16 surrogates"
```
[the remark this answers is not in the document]
xernobyl: Agreed — fixed in 01ff287. The reason now reads `must contain valid UTF-8 and no NULL bytes or unpaired UTF-16 surrogates`, since the `utf8.Valid` gate runs before the escape walk and a caller failing it was being told to remove NUL bytes it may not have had.

tests/feedback_record_metadata_validation_test.go
```diff
@@ -0,0 +1,109 @@
...
+// ENG-2745: metadata that Postgres cannot store must come back as a 400 naming `metadata`, never as
+// an unmapped 500. Two mechanisms are under test, so this deliberately drives the full HTTP stack
+// against the real database: the storable_json request validator (NUL bytes, unpaired surrogates)
+// and the repository's SQLSTATE 22003 mapping (numbers outside numeric range, which no byte scan
+// can see — the value is fifteen bytes of perfectly valid JSON).
+func TestFeedbackRecordMetadataUnstorableInputIsA400(t *testing.T) {
+	server, cleanup := setupTestServer(t)
+	defer cleanup()
+
+	client := &http.Client{}
```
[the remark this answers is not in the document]
xernobyl: Fixed in 01ff287 — 30s timeout on the client. Worth having: the subtests hit a real database, so a blocked handler should fail the run rather than stall it.

internal/repository/feedback_records_repository.go
```diff
@@ -947,6 +962,11 @@ func (r *FeedbackRecordsRepository) Update(
 				return huberrors.NewNotFoundError("feedback record", "feedback record not found")
 			}
 
+			var pgErr *pgconn.PgError
+			if errors.As(scanErr, &pgErr) && pgErr.Code == numericOverflowSQLState {
```
BhagyaAmarasinghe, line 966: **🟠 Major · tests**
The Update-side 22003 mapping has no test — deleting it stays green

- Delete `:966-968` and both `go test ./internal/...` and `make tests` stay green: no test PATCHes a record with `metadata` (`grep -rn MethodPatch tests/`).
- Breaks when: a later edit to `Update`'s error branches drops or reorders it — PATCH 500s again with a green suite. The body claims "both write paths" and "`PATCH` behaves identically"; only Create is tested.
- It does work today — I drove all three classes plus a nested `1e400000` through PATCH on pg18, each 400 naming `metadata`. Fix: one PATCH subtest.

openapi.yaml
```diff
@@ -3264,7 +3282,10 @@ components:
                     pattern: '^[^\x00]*$'
                 metadata:
                     type: object
-                    description: Additional context
+                    description: >-
+                        Arbitrary context stored with this record, returned with equivalent values
+                        (key order and number formatting are normalized). Omitted when the record has
+                        no metadata; a record explicitly stored with a JSON `null` returns `null`.
```
BhagyaAmarasinghe: **🟠 Major · drift**
Response schema forbids the `null` this description promises

- `:3284` is `type: object` in an `openapi: 3.1.0` document, so 2020-12 rejects `null`. `source_id` at `:3070` uses the idiom, `type: [string, "null"]`.
- Breaks when: a record is created with `"metadata": null` — confirmed on pg18 that it stores jsonb `null` and the response carries `"metadata": null`, which a generated client or `response_schema_conformance` rejects. Schemathesis runs `--phases examples`, so CI never generates it.
- Fix: `type: [object, "null"]`, or drop the sentence.

openapi.yaml
```diff
@@ -3047,7 +3047,25 @@ components:
...
+                        Use snake_case keys and keep them stable across records, since a key is what a
+                        dashboard groups by: `customer_tier`, not a mix of `customerTier` and `tier`.
+                        Numbers keep full precision in storage (unlike `value_number`, which is a
+                        float64), so a large integer id round-trips exactly through this API — though
+                        a JavaScript client will still lose precision above 2^53 when it parses the
+                        response.
+
+                        Do not put personal data in metadata unless you mean to retain it — it is not
+                        redacted, and unlike an answer it is duplicated onto every record of a
+                        submission. Rejected with 400, because none of them can be stored: invalid
```
BhagyaAmarasinghe: **🟡 Minor · drift**
The new privacy paragraph is wrong on each of its three claims

- "unless you mean to retain it" — `DELETE /v1/feedback-records?user_id=`, documented at `:424` as Article 17 erasure, deletes whole rows, metadata included. It is retained no more than `value_text`.
- "it is not redacted" implies something else is; `grep -rniE 'redact|anonymi|scrub' internal/` matches nothing.
- "duplicated onto every record of a submission" is a client convention — there is no bulk-create endpoint and nothing in `internal/` copies it across records. This is the paragraph a privacy reviewer reads.

tests/feedback_record_metadata_validation_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+		assert.Equal(t, http.StatusBadRequest, status, payload)
+		assert.Contains(t, metadataParamNames(t, payload), "metadata")
+		assert.Contains(t, payload, "outside the storable range")
+	})
+
+	t.Run("ordinary metadata still stores", func(t *testing.T) {
+		status, payload := postRecord(t, `{"source":"link","big":9007199254740993}`)
+
+		assert.Equal(t, http.StatusCreated, status, payload)
```
BhagyaAmarasinghe: **🟡 Minor · tests**
Nothing pins the exact round-trip the spec now promises

- `:105-109` asserts only `201`; the body is passed as a failure message and never parsed. No test in `tests/` or `internal/` asserts any metadata value comes back.
- `openapi.yaml:3059-3062` newly promises "a large integer id round-trips exactly", and `9007199254740993` is clearly chosen for it.
- Breaks when: a later change reads metadata through `map[string]any` — it returns `...92`, suite green. Fix: parse the 201 body and `assert.JSONEq` on `metadata`.

BhagyaAmarasinghe (commented):
**2 major · 2 minor · 1 trivial · +1 pre-existing** — reviewed at `b0f32516a3`

**Checked**

- `go build`, `go vet ./internal/...`, `go test ./internal/...` — pass
- 12k generated escape-soup docs: scanner verdict == pg18 `::jsonb`, both ways
- Every jsonb numeric-range failure I could produce is SQLSTATE 22003
- PATCH on a real pg18: all three classes 400, naming `metadata`
- Deep nesting: Go's decoder caps below where pg's parser trips
- CodeRabbit's two comments are fixed at this head; not repeated

**Findings**

- **🟠 Major · tests** `internal/repository/feedback_records_repository.go:966` — The Update-side 22003 mapping has no test — deleting it stays green
- **🟠 Major · drift** `openapi.yaml:3288` — Response schema forbids the `null` this description promises
- **🟡 Minor · drift** `openapi.yaml:3066` — The new privacy paragraph is wrong on each of its three claims
- **🟡 Minor · tests** `tests/feedback_record_metadata_validation_test.go:108` — Nothing pins the exact round-trip the spec now promises
- **🔵 Trivial · drift** — The body's before/after quotes a `reason` string the code does not emit
  - Shown as `must not contain NULL bytes or unpaired UTF-16 surrogates`; `internal/api/validation/validation.go:398` emits `must contain valid UTF-8 and no NULL bytes or unpaired UTF-16 surrogates`, which the new tests assert verbatim. Everything else in the block checks out.
- **🟠 Major · correctness · pre-existing** — The same 500 still happens for invalid UTF-8 in a query param or path
  - Nothing validates UTF-8 outside a JSON body — `normalizeIdentifierValue` (`internal/service/id_validation.go:53`) and `no_null_bytes` check only `\x00` — so raw bytes reach pgx. A body string escapes it only because `encoding/json` rewrites to U+FFFD.
  - Reproduced on pg18: `%ed%a0%bd` as `tenant_id`/`source_id`/`user_id` on `GET`+`DELETE /v1/feedback-records`, `/count`, `GET /v1/tenants/{id}/settings` and `GET /v1/webhooks` all 500 with `SQLSTATE 22021`.
  - Follow-up ticket, not this PR. Same for the five `json.RawMessage` fields on `TaxonomyRunResultRequest` (`internal/models/taxonomy.go:225-255`), which reach jsonb with neither the tag nor a 22003 mapping.

_Reply per finding: the fix, why it's wrong, or a ticket._

_claude-opus-5 · high_

internal/repository/feedback_records_repository.go
xernobyl: Right, and worse than unbound — I had actually written this subtest, then destroyed it with a `git checkout --` while mutation-testing something else, and did not notice until re-reviewing. Twice now; noted.

Added in b077d65. It PATCHes all three classes, with two guards against passing for the wrong reason: an empty-PATCH baseline first (asserting 200, so a 400 below cannot be "any body fails"), and `metadataParamNames` asserting the reason names `metadata` — a bad id would name `id`.

Mutation-verified your exact scenario: deleting `:966-968` now fails `the_update_path_rejects_the_same_three_classes`, and so does dropping the Update struct's validator tag.

openapi.yaml
xernobyl: Confirmed and fixed in b077d65 — I POSTed `"metadata": null` on pg18 and got a 201 with `"metadata": null` back, so the schema was forbidding what the server does.

Two things beyond what you flagged, both the same finding one field over:

- the **PATCH request body** at `:3665` was a third `metadata` schema with the identical drift — also `[object, "null"]` now
- the response schema is now **untyped** rather than `[object, "null"]`. Applying your rule properly, null is not the only shape it forbids: `42`, `[1,2,3]`, `"str"` and `true` all return 201 and round-trip verbatim today, so an object-only response schema would make a generated client reject a legitimate record. The description says an object is the supported shape and what the request schemas accept.

Request schemas stay `[object, "null"]` deliberately — narrower than the server on a request is the safe direction, and the v3 gateway already enforces object shape.

openapi.yaml
xernobyl: All three claims were wrong, and the erasure one was wrong in the dangerous direction. Rewritten in b077d65.

Your point about `DELETE` went further than the original text: I had written the erasure sentence unconditionally, but the delete matches on `user_id`, so a row written without one cannot be reached at all — the old wording was over-broad and failed safe, mine failed open. It now names the precondition and tells the caller to set `user_id` on anything they may have to erase.

Also dropped the "duplicated onto every record" claim (a Formbricks-pipeline behaviour, not a Hub one), and flipped the `value_text` analogy: metadata is now *contrasted* with it, since only `value_text` is shipped to LLM and embedding providers.

tests/feedback_record_metadata_validation_test.go
xernobyl: Fixed in b077d65 — and thanks for spotting that the obvious fix would not have worked: `assert.JSONEq` unmarshals both sides to `interface{}`, so it would lose the precision on both and pass regardless.

It now decodes with `json.RawMessage` plus `json.Decoder.UseNumber()` and asserts `json.Number("9007199254740993")`. Mutation-verified: removing the `UseNumber()` line turns it red with `expected: json.Number("9007199254740993")`.

openapi.yaml
```diff
@@ -3263,8 +3286,16 @@ components:
...
                 metadata:
-                    type: object
-                    description: Additional context
+                    description: >-
+                        Arbitrary context stored with this record, returned with equivalent values
+                        (key order is normalized, and so are number exponents — `1e2` reads back as
+                        `100`). Omitted when the record has no metadata; a record explicitly stored
+                        with a JSON `null` returns `null`.
+
+                        Deliberately untyped: an object is the supported shape and what the request
```
BhagyaAmarasinghe: **[P1] Preserve a concrete response type for `metadata`**

Leaving this schema typeless introduces a new Stainless `Schema/TypeMissing` diagnostic and changes the generated TypeScript response from `metadata?: { [key: string]: unknown }` to `metadata?: unknown`. Existing SDK consumers that inspect metadata will stop type-checking after an otherwise patch-level fix. Please represent the legacy shapes explicitly (for example, a JSON-value union covering object, array, string, number, boolean, and null) so scalar rows remain valid without erasing the public response type.

openapi.yaml
```diff
@@ -3046,8 +3046,31 @@ components:
...
+                        booleans, null, or nested objects and arrays.
+
+                        Use snake_case keys and keep them stable across records, since a key is what a
+                        dashboard groups by: `customer_tier`, not a mix of `customerTier` and `tier`.
+                        Numbers keep full precision in storage (unlike `value_number`, which is a
+                        float64), so a large integer id round-trips exactly through this API — though
+                        a JavaScript client will still lose precision above 2^53 when it parses the
+                        response.
+
+                        Nothing here is redacted or sent anywhere else — unlike `value_text`, which
```
BhagyaAmarasinghe: **[P2] Do not promise that metadata is never sent elsewhere**

Create and update publish the full `FeedbackRecord`, and `publicWebhookData` forwards non-delete event data unchanged, so `feedback_record.created` and `feedback_record.updated` webhooks include `metadata` at the configured webhook URL. This sentence could encourage callers to store sensitive data under a false guarantee. Please say that metadata is not sent to the LLM/embedding enrichment providers, while noting that it is included in feedback-record webhooks.

BhagyaAmarasinghe (changes requested):
Requesting changes for two API-contract issues at the current head. The validator and repository mappings look correct, and the targeted POST/PATCH regression coverage passes.

openapi.yaml
xernobyl: Fair, and the tradeoff was the wrong way round — I was protecting a hypothetical strict validator and broke a real generated type to do it.

Fixed in f546f54: `type: [object, array, string, number, boolean, "null"]`, so the array and scalar rows the column actually holds still validate while the response keeps a concrete public type. Spectral then requires an `items` sibling once `array` is in the union, so that is there as `items: {}` with a note on why it is unconstrained.

The description now says plainly that an object is the supported shape and the only one the request schemas accept, and that the wider union exists for reading legacy rows — so it documents the intent rather than implying scalars are a supported way to write metadata.

openapi.yaml
xernobyl: You are right, and this is the second time in this PR my privacy prose failed open rather than safe — worth me being more careful with guarantee-shaped sentences.

Verified before changing it: `publicWebhookData` (`internal/service/webhook_payload.go:41-52`) returns the event data unchanged for anything that is not a delete event, so `feedback_record.created` and `feedback_record.updated` both carry the full record, metadata included, to the configured URL.

Fixed in f546f54. It now separates the two: **not** sent to the LLM or embedding enrichment providers, unlike `value_text` — but **is** included in those two webhook payloads, so it reaches whatever URL a tenant has configured.

BhagyaAmarasinghe (approved):

PR #135, opened 2026-09-14
chore(deps): update Go, modules, tooling, CI actions and docs dependencies to current

xernobyl:
## What does this PR do?

Brings the Hub up to current across the board: the Go toolchain and every module, the pinned CLI tools, the container base images, all 14 pinned GitHub Actions, and the `docs/` npm tree.

Dependabot had **16 open alerts** on this repo and no `dependabot.yml` to raise PRs for them, so nothing was clearing them. This closes all 16.

Resolves [ENG-3213](https://linear.app/formbricks/issue/ENG-3213/hub-update-go-modules-tooling-ci-actions-and-docs-dependencies-to).

### Alerts cleared

`go.mod` — all three are grpc, fixed by landing on **1.83.2**:

| Advisory | Severity | Range |
|---|---|---|
| GHSA-2v4p-qf9q-27wj | high | `< 1.82.2` |
| GHSA-vp52-pcj8-j9qc | high | `<= 1.83.0` |
| GHSA-qc2q-p7wx-3px3 | medium | `<= 1.83.0` |

`docs/` — ten alerts, including the only critical one:

| Advisory | Severity | Package | Now |
|---|---|---|---|
| GHSA-26w7-cxv4-gfx2 | **critical** | astro | 7.3.2 |
| GHSA-f88m-g3jw-g9cj, GHSA-rgj7-g3m4-5g8c | high | sharp | 0.35.4 |
| GHSA-f65p-4m7j-42xc, GHSA-jqff-g426-hqxp, GHSA-fph4-wmhf-6fwf, GHSA-5jgf-p345-68v8 | high | fast-uri | 3.1.7 |
| GHSA-2883-xcg3-v3hh | high | js-yaml | 4.3.2 |
| GHSA-hmw2-7cc7-3qxx | high | form-data | 4.0.6 |
| GHSA-w27v-7q3p-w38r | high | svgo | 4.1.0 |
| GHSA-4vpr-x523-8j87 | medium | svgo | 4.1.0 |

### Versions

| | From | To |
|---|---|---|
| Go | 1.26.6 | 1.27.1 |
| alpine (runtime) | 3.21 | 3.24 |
| goose | v3.27.1 | v3.28.0 |
| river | v0.39.0 | v0.47.0 |
| golangci-lint | v2.11.4 | v2.13.2 |
| govulncheck | v1.3.0 | v1.8.0 |
| schemathesis | 4.4.4 | 4.27.1 |
| Python (CI) | 3.13 | 3.14 |
| pnpm | 11.5.0 | 12.4.1 |
| astro | 7.2.4 | 7.3.2 |
| @astrojs/starlight | 0.41.7 | 0.42.0 |

Plus every remaining Go module and all 14 pinned actions. Each new action SHA was verified to be the commit its version comment claims.

### The parts that needed a judgement call

**TypeScript stops at 6.0.3, not the latest 7.0.2.** `@astrojs/check` peers `^5 || ^6`. TypeScript 7 does not yet expose the programmatic API Volar needs to embed it for `.astro` files, so `astro check` cannot run against it. That API is expected in 7.1.

**posthog-js keeps its `^1.376.0` range.** pnpm 12 enforces a supply-chain delay on freshly published packages; posthog cut 1.431.6 this morning, so the policy resolves 1.430.3 instead. `pnpm install` auto-wrote a `minimumReleaseAgeExclude` entry to bypass the hold — I removed it. Skipping the cooldown on a browser analytics SDK is precisely what the policy exists to prevent.

**form-data needed an override.** `httpsnippet`, via `starlight-openapi`, pins it to exactly `4.0.4` and upstream has not re-released. The override forces the patched 4.0.6, mirroring what the Dockerfile already does for the Go CLIs.

**satteri moved to 0.10.5.** Upstream never published a `darwin-arm64` build for 0.10.4, the version the old lockfile resolved — `astro check` could not run on Apple Silicon at all. This is pre-existing on `main`, fixed here because the lockfile was being regenerated anyway.

**pnpm/action-setup had to go to v6.1.0.** That is the release that added pnpm v12 support; with `packageManager` now at 12.4.1, v4.3.0 would have failed the docs build outright.

**River 0.40 introduced schema version 7** (and dropped the unused `river_client` / `river_client_queue` tables). **No deploy action is required** — the chart's migration Job and compose's `hub-migrate` both run `river migrate-up` on every install/upgrade. I confirmed migration 007 applies cleanly to a database sitting at version 6.

**Go 1.27 backs `encoding/json` with the v2 implementation**, which now rejects duplicate object keys and invalid UTF-8. This is the one behaviour change with real reach for a JSON API, which is why the full integration and contract suites were run rather than just the unit tests. Nothing depended on the old leniency.

### Deliberately not included

- `node-version` stays on **24**, the active LTS.
- The TEI embeddings image (`cpu-1.9` in the chart) — production inference with tuned args, owned by ENG-2791 / ENG-2798.
- No `dependabot.yml`. Worth adding so this does not accumulate again, but it is a separate change.

## How should this be tested?

Everything below was run on this branch. A reviewer can reproduce with a pgvector pg18 database and `DATABASE_URL` pointed at it.

**Go**

```bash
make build && make fmt && make lint && make test-unit && make tests && make migrate-validate
```

- `make build` — clean
- `make lint` (golangci-lint **2.13.2**) — `0 issues`
- `make test-unit` — 17 packages pass
- `make tests` — 377 integration tests pass against pgvector pg18
- `make migrate-validate` — clean
- `go vet ./...` — clean
- `govulncheck ./...` (**1.8.0**) — *"Your code is affected by 0 vulnerabilities."* The one remaining module-level finding is `golang.org/x/crypto/openpgp` being unmaintained (GO-2026-5932, no fix available, not called by Hub).

The linter bump surfaced 86 pre-existing findings. 43 of them came from a single gofumpt rule and were dropped by configuring it off (see below); the rest were machine-applied with `golangci-lint --fix`, leaving 19 files of genuine Go 1.27 modernization: `errors.As` -> `errors.AsType`, `reflect.Ptr` -> `reflect.Pointer`, `strings.SplitN(_, _, 2)` -> `strings.Cut`, and one `reflect.TypeAssert`. That is isolated in its own commit so the dependency diff stays readable.

**On `.golangci.yml`.** golangci-lint deprecates gofumpt's `extra-rules: true` and its notice says to use `extra.group-params` instead. That is not an equivalent swap — `extra` has three sub-options (`group-params`, `clothe-returns`, `balance-calls`) and `extra-rules: true` enabled all three, so following the notice literally would have silently switched two of them off. The replacement is therefore spelled out explicitly, with `balance-calls` deliberately **off**: it moves the closing paren of every wrapped call onto its own line, which reformatted ~43 call sites for no benefit. The repo has never been formatted that way — the gofumpt bundled with 2.11.4 did not apply that rule — so leaving it off preserves the existing house style instead of churning it. `make fmt` is idempotent against the committed tree under this config, which is what the Code Quality job asserts.

**Migrations**

```bash
goose -dir migrations postgres "$DATABASE_URL" up   # -> version 26
river migrate-up --database-url "$DATABASE_URL"      # -> applies 007
```

Both run clean; River 007 applies against a schema at version 6.

**Container**

```bash
docker build -t hub-deps-check:local .
```

Builds. Verified inside the image: alpine `3.24.1`, `goose v3.28.0`, `river v0.47.0` *"Built with go1.27.1"*, and all three binaries present. Inspecting the shipped binary confirms the fix reaches the artifact, not just `go.mod`:

```
$ go version -m /build/bin/hub-api | grep grpc
    dep    google.golang.org/grpc    v1.83.2
```

**API contract tests**

Both Schemathesis invocations from `api-contract-tests.yml` were run locally against a live `hub-api` on the upgraded stack, because 4.x reworked operation filtering into `--include-TYPE` / `--exclude-TYPE` and I wanted to confirm the existing flags still resolve:

- Step 1: 28/32 operations selected (the four exclusions applied), **61 cases generated, 61 passed**
- Step 2: 4/4 selected, **6 cases generated, 6 passed**

**Docs**

```bash
cd docs && pnpm install --frozen-lockfile && pnpm check && pnpm build && pnpm check:links
```

- `pnpm check` — 0 errors, 0 warnings, 0 hints
- `pnpm build` — 58 pages, sharp image optimization working
- `pnpm check:links` — 2450 internal references, none broken
- `pnpm install --frozen-lockfile` — clean, which is what CI runs

Starlight 0.42 rebuilt the mobile menu on the popover API and removed `<starlight-menu-button>`, the `aria-expanded` attribute and `body[data-mobile-menu-expanded]`. `theme.css` touches none of them and there are no component overrides, so nothing here depends on the removed markup. I rendered the built site at desktop and at 375px to confirm, including opening the mobile menu:

| Desktop | Mobile menu open |
|---|---|
| Sidebar, theme colours and active-item styling intact | New `sl-sidebar-pane` popover opens; custom teal accents and active-item background still applied |

### Required

- [x] Ran `make build`
- [x] If database schema changed: no goose migration in this PR. River's own schema moves to version 7, applied by the existing `river migrate-up` step in the chart's migration Job. `make migrate-validate` passes.

### Appreciated

- [x] If API changed: API surface is unchanged; contract tests were run anyway because the toolchain moved (results above)
- [ ] If API behavior changed: n/a — no behaviour change
- [x] Updated docs in `docs/` if changes were necessary — dependency-only; content unchanged, site builds and renders
- [ ] Ran `make tests-coverage` — no logic changes; the only hand-written source change is one added enum case

## Note on the one non-mechanical source change

`genai` 1.71.0 adds `FinishReasonTooManyToolCalls`, which the `exhaustive` linter flagged in `internal/googleai/client.go`. It joins the retryable list beside the other tool-call reasons: this client makes no tool calls, so the reason cannot occur, and the file's existing rule is to enumerate reasons explicitly rather than let them fall through to the default.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mattinannt (approved):
thank you for the quick fix 😊

